PR #65, opened 2022-01-13
feat: asset publishing no longer depends on asset hash

kaizen3031593:
The previous mechanism for publishing assets depended on the stability of the asset hash between the local `cdk synth` and the Synthesize step in github actions. Long story short, this asset hash is actually quite brittle and can break due to differing versions of node. The current workaround is to remove CDK Metadata from the template, since that is the resource that encodes the node version.

This PR aims to remove asset hashes from `deploy.yaml` entirely, so as to remove the dependency to this hash. During `cdk synth`, we now write a bash script per asset to `cdk.out`, and reference this script in the github step. We will now always find the asset we are looking for.

A side-effect is that `deploy.yaml` no longer changes when the asset changes, since we don't depend on the hash.

BREAKING CHANGE: The publish asset step executes a script from `cdk.out` rather than call `cdk-assets` directly.


src/pipeline.ts
```diff
@@ -220,14 +220,24 @@ export class GitHubWorkflow extends PipelineBase {
 
   private jobForAssetPublish(node: AGraphNode, assets: StackAsset[], options: Context): Job {
     const installSuffix = this.cdkCliVersion ? `@${this.cdkCliVersion}` : '';
-    const relativeToAssembly = (p: string) => path.posix.join(cdkoutDir, path.relative(options.assemblyPath, p));
+    const pathDirectories = this.workflowPath.split(path.sep);
+    const cdkoutDir = path.join(pathDirectories.slice(0, pathDirectories.indexOf('.github')).join(path.sep), 'cdk.out');
```
kaizen3031593, line 224: is this too brittle? We already ensure that `this.workflowPath` includes `.github/workflows` earlier, so we are confident that `pathDirectories.indexOf('.github')` succeeds.

But all of this is to ensure that the tests, which have a workflow path of `some/temp/dir/.github/workflows/deploy.yml`, have their asset scripts stored in `some/temp/dir/cdk.out`. Otherwise, these assets get put in the source directory, which is weird.

test/__snapshots__/github.test.ts.snap
```diff
@@ -398,7 +358,7 @@ jobs:
         uses: actions/download-artifact@v2
         with:
           name: cdk.out
-          path: cdk.out
+          path: /var/folders/5t/qdsk26vd29l6bk_3kg6ydwx00000gs/T/cdk-pipelines-github-WmeGA4/cdk.out
```
kaizen3031593: This diff relates to the issue I've raised above. I don't think we want `path: cdk.out` anymore since that points to the base directory, not the temp directory. This previously wasn't an issue because we didn't actually create + add any files to `cdk.out`.

test/__snapshots__/github.test.ts.snap
```diff
@@ -370,7 +334,7 @@ jobs:
         uses: actions/download-artifact@v2
         with:
           name: cdk.out
-          path: cdk.out
+          path: temp/cdk.out
```
kaizen3031593, line 337: This diff relates to the issue I've raised above. I don't think we want path: cdk.out anymore since that points to the base directory, not the temp directory. This previously wasn't an issue because we didn't actually create + add any files to cdk.out.

rix0rrr (approved):

PR #81, opened 2022-01-27
feat: github openId construct and ability to authenticate to aws via openId 

kaizen3031593:
This PR adds an openId option for authenticating against AWS. Previously the only way to authenticate against AWS was to provide long-lived github secrets with AWS credentials. Now we can configure a role that uses an oidc identity provider to authenticate.

This PR adds two functionalities:

A CDK construct `GitHubActionRole` that configures the necessary role for you, and returns the role arn.

A new property in `GitHubWorkflow` called `gitHubActionRoleArn` that takes in the role arn and if set, the workflow will authenticate via role rather than github secrets.

The role that we create via `GitHubActionRole` has the necessary permissions to assume CDK bootstrapped roles and login to ecr.


src/oidc-provider.ts
```diff
@@ -0,0 +1,93 @@
...
+          [`${rawEndpoint}:sub`]: `repo:${props.username}/${props.repository}:ref:refs/heads/${props.branch ?? 'main'}`,
+        },
+      },
+      'sts:AssumeRoleWithWebIdentity',
+    );
+
+    const role = new iam.Role(this, 'github-role', {
+      roleName: 'GithubActionRole',
+      assumedBy: principal,
+      managedPolicies: [iam.ManagedPolicy.fromAwsManagedPolicyName('AdministratorAccess')],
```
kaizen3031593, line 88: I assume this is not okay but I don't know what else to do here. How will we know what permissions to grant this role? Or is it okay to have admin access because the way it is configured, only github actions from your github repo/branch can use it.

src/oidc-provider.ts
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * You can then utilize the role arn as a stack output and send it into the Github Workflow app via
+ * the `githubOidcRoleArn` property. The role arn will be `arn:aws:iam::<accountId>:role/GithubActionRole`.
+ *
+ * @see https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services
+ */
+export class GithubOidc extends Construct {
+  /**
+   * The arn of the role that gets created. The arn will equal
+   * `arn:aws:iam::<accountId>:role/GithubActionRole`.
```
kaizen3031593, line 54: I feel like the best we can do is just advertise the role arn like this, because we do not have knowledge of the account id in github workflows. We have them at the stage/stack level but I don't think that helps here.

kaizen3031593:
TODO: update README when the API is more settled

src/oidc-provider.ts
rix0rrr: For assuming roles, I think it needs the following:

* Can assume ANY role in a different account
* Can assume ONLY bootstrap roles in the same account

Roughly like what we're doing here for S3: https://github.com/aws/aws-cdk/blob/master/packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml#L409

Roughly this?

```yaml
- Sid: AssumeBootstrapRoles
  Effect: Allow
  Action:
    - sts:AssumeRole
  Resource: "*"
  "Condition": {
    "ForAnyValue:StringEquals": {
      "iam:ResourceTag/aws-cdk:bootstrap-role": ["deploy", "lookup", "file-publishing", "image-publishing"]
    }
  }
```

src/pipeline.ts
```diff
@@ -127,7 +146,7 @@ export class GitHubWorkflow extends PipelineBase {
   protected doBuildPipeline() {
     const app = Stage.of(this);
     if (!app) { throw new Error('');}
```
rix0rrr, line 148: needs an error ?

rix0rrr (changes requested):
Good work! Needs a README update as well btw, walk people through how to use this

src/oidc-provider.ts
```diff
@@ -0,0 +1,93 @@
...
+import { Construct } from 'constructs';
+
+/**
+ * Properties for the GithubOidc construct.
+ */
+export interface GithubOidcProps {
+  /**
+   * Your Github username.
+   */
+  readonly username: string;
```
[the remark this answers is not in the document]
rix0rrr: And potentially make this a list?

src/oidc-provider.ts
kaizen3031593: both assume role and ecr permissions added.

src/oidc-provider.ts
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Properties for the AwsOidc construct.
+ */
+export interface AwsOidcProps {
+  /**
+   * Your Github username and repository passed in as a single string.
+   * For example, `owner/repo`.
+   */
+  readonly repoString: string;
```
rix0rrr, line 13: Just `repo` I think.

src/pipeline.ts
```diff
@@ -136,8 +156,10 @@ export class GitHubWorkflow extends PipelineBase {
 
   protected doBuildPipeline() {
     const app = Stage.of(this);
-    if (!app) { throw new Error('');}
-    const cdkoutDir = app?.outdir;
+    if (!app) {
+      throw new Error('The Github Workflow must extend App');
```
rix0rrr, line 160: Not extend, rather something like "be defined in the scope of".

src/pipeline.ts
```diff
@@ -487,30 +513,41 @@ export class GitHubWorkflow extends PipelineBase {
...
+      }));
+
+      if (assumeRoleArn) {
+        // Result of initial credentials with OIDC role are these environment variables
+        steps.push(awsCredentialStep('Assume CDK Deploy Role', {
+          region,
+          accessKeyId: '${{ env.AWS_ACCESS_KEY_ID }}',
+          secretAccessKey: '${{ env.AWS_SECRET_ACCESS_KEY }}',
+          sessionToken: '${{ env.AWS_SESSION_TOKEN }}',
+          roleToAssume: getDeployRole(assumeRoleArn),
```
rix0rrr, line 536: ExternalId?

src/pipeline.ts
```diff
@@ -487,30 +513,41 @@ export class GitHubWorkflow extends PipelineBase {
...
+        }));
+      }
+    } else {
+      steps.push(awsCredentialStep('Authenticate Via GitHub Secrets', {
+        region,
+        accessKeyId: `\${{ secrets.${this.awsCredentials.accessKeyId} }}`,
+        secretAccessKey: `\${{ secrets.${this.awsCredentials.secretAccessKey} }}`,
+        sessionToken: `\${{ secrets.${this.awsCredentials.sessionToken} }}`,
+        roleToAssume: assumeRoleArn,
+        roleExternalId: assumeRoleArn ? 'Pipeline' : undefined,
```
rix0rrr, line 546: Why a hardcoded externalId here btw?

rix0rrr (changes requested):
Approved modulo some small tweaks

src/pipeline.ts
kaizen3031593: talked about not using externalID anywhere because it does not add security for our use case

src/oidc-provider.ts
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   * Reference an existing github actions provider.
+   * You do not need to pass in an arn because the arn for such
+   * a provider is always the same.
+   */
+  public static existingGithubActionsProvider(scope: Construct): iam.IOpenIdConnectProvider {
+    return iam.OpenIdConnectProvider.fromOpenIdConnectProviderArn(
+      scope,
+      'GithubActionProvider',
+      `arn:aws:iam::${Aws.ACCOUNT_ID}:oidc-provider/token.actions.githubusercontent.com`,
```
kaizen3031593, line 61: I did this, but am curious about a use case where the user is building this construct in an account different from the one they want the github action role to be in (and thus, `Aws.ACCOUNT_ID` will fail).

rix0rrr (approved):

PR #83, opened 2022-01-28
feat: support docker login

kaizen3031593:
Adds the ability to login to dockerHub, ecr, or custom repositories.


src/docker-credentials.ts
```diff
@@ -0,0 +1,71 @@
+export class DockerCredential {
+  /**
+   * Create a credential for DockerHub. This method assumes that you will have long-lived
+   * Github Secrets stored under the usernameKey and personalAccessTokenKey.
```
rix0rrr, line 4: Mention the defaults.

src/docker-credentials.ts
```diff
@@ -0,0 +1,71 @@
...
+   */
+  public static dockerHub(creds: DockerHubCredentialSecrets = {}): DockerCredential {
+    return new DockerCredential(
+      'docker',
+      undefined,
+      creds.usernameKey ?? 'DOCKERHUB_USERNAME',
```
rix0rrr, line 10: Can you double-check these names are commonly used across the internet?

src/docker-credentials.ts
```diff
@@ -0,0 +1,71 @@
...
+      creds.personalAccessTokenKey ?? 'DOCKERHUB_TOKEN',
+    );
+  }
+
+  /**
+   * Create a credential for ECR. This method will reuse your AWS credentials to log in to AWS.
```
rix0rrr, line 16: Reiterate what you mean when you say "your AWS credentials".

src/docker-credentials.ts
```diff
@@ -0,0 +1,71 @@
...
+   * @default 'DOCKERHUB_USERNAME'
+   */
+  readonly usernameKey?: string;
+
+  /**
+   * The key of the Github Secret containing the DockerHub personal access token.
+   *
+   * @default 'DOCKERHUB_TOKEN'
+   */
+  readonly personalAccessTokenKey?: string;
```
rix0rrr, line 58: Is this the name DockerHub name?

src/docker-credentials.ts
rix0rrr: Probably also add a description of the IAM permissions this user/role needs.

rix0rrr (commented):
Conditionally approved, but I'm not sure that exists on this repo so not approving yet 😝

src/docker-credentials.ts
kaizen3031593: This looks to be the correct defaults. It says so on the official [docker login action](https://github.com/docker/login-action#docker-hub), and comes up often when I search for them on google.

src/docker-credentials.ts
kaizen3031593: we create personal access tokens in DockerHub that is what this value references. I'm asking the user to store the key to the github secret that holds the personal access token, so I think this is an appropriate name.

rix0rrr (approved):

PR #91, opened 2022-02-04
chore: rewrite readme

kaizen3031593:
Rewriting the README story so that it reads more similarly to other CDK modules.


README.md
```diff
@@ -55,8 +69,222 @@ documentation for more details.
...
 * The workflow expects the GitHub repository to include secrets with AWS
   credentials (`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`).
+  
+## AWS Credentials
+
+There are two ways to supply AWS credentials to the workflow. The default is to
+store long-lived AWS user credentials in GitHub secrets. Alternatively, you can
+use an IAM role that trusts the GitHub OpenId Connect provider to authenticate.
+Using OIDC to authenticate is not the default because it is a newer option, but
+there are compelling benefits to this approach. You can read more
```
kaizen3031593, line 79: How detailed should I be here? Should I even be pushing OIDC as a better alternative or just be unopinionated here?

README.md
```diff
@@ -2,10 +2,24 @@
 
 ![Experimental](https://img.shields.io/badge/experimental-important.svg?style=for-the-badge)
 
-NOTICE: this library is still not published to package managers. Stay tuned.
```
kaizen3031593, line 5: It looks like we are already publishing to npm and PyPI at least

README.md
```diff
@@ -2,10 +2,24 @@
...
-
 Deploy CDK applications through GitHub workflows.
```
rix0rrr, line 5: Maybe add a pointer here to CDK Pipelines as well, and say "THIS is like THAT, except..." ?

rix0rrr: Also like two more lines about the purpose would be useful. Your target audience is someone with passable knowledge about computers, software systems, GitHub and AWS, but not necessarily all the intricacies of deploying, how to setup their CDK app, and what other solutions are available on the market.

README.md
rix0rrr: Definitely push OIDC as the "better solution", but don't say "because it's a newer option".

Instead say something like "it requires preparing your AWS account beforehand" or something. Also, I find that in technical writing:

* I have opinions
* I like bullet points (recommended)

The bullet points make a list of alternatives easily scannable. You can use the text blocker *after* the bullet points to elaborate on the differences between them (if the text goes on for more than 2 or 3 sentences per bullet, you might consider splitting up into a paragraph per bullet for elaboration).

### Sections

If there's even more text than 4-5 sentences, and/or you want to include an example, **sections** are the next go-to after paragraphs.

README.md
```diff
@@ -55,8 +69,222 @@ documentation for more details.
...
+  synth: new ShellStep('Build', {
+    commands: [
+      'yarn install',
+      'yarn build',
+    ],
+  }),
+  awsCredentials: {
+    accessKeyId: 'MY_ID',
+    secretAccessKey: 'MY_KEY',
+    sessionToken: 'MY_TOKEN', // default is no token
```
rix0rrr, line 107: Tokens don't make sense here, as long-lived credentials never have tokens.

Leave the parameter in, because who knows what an enterprising user might use these for. But there's not really a point to wasting valuable brain space for doc readers on including this field.

README.md
```diff
@@ -55,8 +69,222 @@ documentation for more details.
...
+```ts
+import { AwsOidc } from 'cdk-pipelines-github';
+import { App, Construct, Stack, StackProps } from 'aws-cdk-lib';
+
+class MyAwsOidcRole extends Stack {
+  constructor(scope: Construct, id: string, props?: StackProps) {
+    super(scope, id, props);
+
+    const provider = new AwsOidc(this, 'oidc-role', {
+      repoString: 'myUser/myRepo',
```
rix0rrr, line 158: Also I think we said this should be a list.

README.md
```diff
@@ -55,8 +69,222 @@ documentation for more details.
...
+
+```ts
+import { AwsOidc } from 'cdk-pipelines-github';
+import { App, Construct, Stack, StackProps } from 'aws-cdk-lib';
+
+class MyAwsOidcRole extends Stack {
+  constructor(scope: Construct, id: string, props?: StackProps) {
+    super(scope, id, props);
+
+    const provider = new AwsOidc(this, 'oidc-role', {
```
rix0rrr, line 157: Maybe call this `AwsOidcRole` then.

rix0rrr: Or honestly--how about `GitHubActionRole` ?

README.md
```diff
@@ -55,8 +69,222 @@ documentation for more details.
...
+      repoString: 'myUser/myRepo',
+      provider: iam.OpenIdConnectProvider.fromOpenIdConnectProviderArn(
+        this,
+        'github',
+        'arn:aws:iam::<account-id>:oidc-provider/token.actions.githubusercontent.com',
```
rix0rrr, line 184: If this is always the same: you can make a static function on the class to retrieve it. Use `Aws.ACCOUNT_ID` to make it use the "current account". We can then write this:

```ts
 const provider = new AwsOidc(this, 'oidc-role', {
      repoString: 'myUser/myRepo',
      provider: AwsOidc.existingGitHubActionsProvider(),
  });
```

Pretty nice.

README.md
```diff
@@ -55,8 +69,222 @@ documentation for more details.
...
+      ),
+    });
+  }
+}
+```
+
+#### Manual Setup
+
+In a nutshell, the IAM role you provide must reference the GitHub OIDC identity
```
rix0rrr, line 193: "In a nutshell" -> think this reads just as well without this bit

README.md
```diff
@@ -55,8 +69,222 @@ documentation for more details.
...
+the IAM role and your GitHub repository. For a step-by-step tutorial on how to
+set this up, see
+[Configuring OpenID Connect in AWS](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services).
+
+In addition to setting up a relationship between GitHub and AWS, the IAM role 
+must also have permissions to assume CDK bootstrapped IAM roles and permissions 
+to access ECR repositories (if you plan on referencing Docker assets in your 
+workflow).
+
+Here is a minimum set of permissions for the IAM role:
```
rix0rrr, line 204: It's great that you have this, but I would split this out into a separate doc maybe? I feel like it clutters the README a bit. At this point, 80% of the doc is on authentication, and barely anything is about the API or how you actually deploy something.

README.md
nikovirtala: Your official GitHub Actions repository recommends using OIDC and short-lived credentials: ["We recommend using GitHub's OIDC provider to get short-lived credentials needed for your actions."](https://github.com/aws-actions/configure-aws-credentials#assuming-a-role)

As a user and an advocate, I'd be happy to see it here too.

README.md
```diff
@@ -55,8 +69,222 @@ documentation for more details.
...
+      "Resource": "*"
+    }
+  ]
+}
+```
+
+### Using Docker in the Pipeline
+
+You can use Docker in GitHub Workflows in a similar fashion to CDK Pipelines,
+except you cannot run Docker in the self-mutate project because there is no
```
rix0rrr, line 237: If it doesn't apply, better to leave it out than confuse the user. They don't need to know about self-mutation when using this. Don't tell them there's a thing they don't need to know about! 😀

rix0rrr (changes requested):

README.md
kaizen3031593: I've reworked this section, let me know if more detail is needed. The CDK Pipeline module starts off the same way, so I thought I'd pay homage

README.md
kaizen3031593: Do you mean the manual set up, or authenticating to AWS as a whole? For now, I feel like authenticating is pretty important and deserves to be in the readme, so I moved the manual set up elsewhere.

PR #96, opened 2022-02-08
feat: remove asset hash from templates

kaizen3031593:
A previous change removed hard-coded asset hashes from the `deploy.yml` file to circumvent brittleness surrounding these hashes. However, they were still hard-coded in the templates we reference. This PR addresses this oversight by removing those as well in favor of referencing job outputs.

Snapshot tests have been updated. There's not really anything more to test here.

BREAKING CHANGE: the hard-coded asset hash is removed from the template string we send into cloudformation.


src/pipeline.ts
```diff
@@ -320,9 +331,17 @@ export class GitHubWorkflow extends PipelineBase {
       });
     };
 
+    const replaceAssetHash = (template: string) => {
+      const hash = path.parse(template.split('/').pop() ?? '').name;
```
kaizen3031593, line 335: this is fairly brittle; is it ok? are there alternatives that I can explore here?

kaizen3031593 (commented):
No tests yet, just a quick POC that I made sure works with the integration test

rix0rrr:
> BREAKING CHANGE: the hard-coded asset hash is removed from the template string we send into cloudformation.

I don't think this is a breaking change. Yes, something will change (that is mostly inevitable). But a breaking change occurs in the following case:  there was some scenario that used to work, but now doesn't work anymore.

Is that happening here? I don't think so.

src/pipeline.ts
rix0rrr: I bet there are cases where it's going to fail, but we'll deal with those when we come to them.

src/pipeline.ts
```diff
@@ -266,27 +268,36 @@ export class GitHubWorkflow extends PipelineBase {
     // create one file and make one step
     const relativeToAssembly = (p: string) => path.posix.join(cdkoutDir, path.relative(path.resolve(cdkoutDir), p));
     const fileContents: string[] = ['set -x'].concat(assets.map((asset) => {
-      return `npx cdk-assets --path "${relativeToAssembly(asset.assetManifestPath)}" --verbose publish "${asset.assetSelector}"`;
+      // we need the jobId to reference the outputs later
+      this.assetHashMap[asset.assetId] = jobId;
+      return [
+        `npx cdk-assets --path "${relativeToAssembly(asset.assetManifestPath)}" --verbose publish "${asset.assetSelector}"`,
+        `echo '::set-output name=hash::${asset.assetId}'`,
```
rix0rrr, line 275: This might be fine, it kinda depends on the multiplicity of things. Does this still work if there are multiple assets per asset step? Maybe that never happens?

src/pipeline.ts
```diff
@@ -320,9 +331,17 @@ export class GitHubWorkflow extends PipelineBase {
...
+      if (this.assetHashMap[hash] === undefined) {
+        throw new Error(`Template asset hash ${hash} not found.`);
+      }
+      return template.replace(hash, `\${{ needs.${this.assetHashMap[hash]}.outputs.asset-hash }}`);
+    };
+
     const params: Record<string, any> = {
       'name': stack.stackName,
-      'template': resolve(stack.templateUrl),
+      'template': replaceAssetHash(resolve(stack.templateUrl)),
```
rix0rrr, line 344: I think what I imagined (and I think we discussed) this code could look like, is something like:

```ts
const outputName = sanitizeName(`a_${stack.artifactId}`);
console.log(`::set-output name=${ouputName}::${stack.templateUrl}`);
const params: Record<string, any> = {
  'template': `\${{ synth.outputs.${outputName} }}`,
};
```

What do you think about this?

kaizen3031593:
> there was some scenario that used to work, but now doesn't work anymore.

I thought a breaking change also happens if someone's CDK template changes without any code changes. Since the executable we are creating here is `deploy.yml`, can't the same argument be made?

Users are going to see a change in their `deploy.yml` file without changing their code. Isn't that breaking?

src/pipeline.ts
kaizen3031593: when there are multiple assets per step, it is because the same asset is getting uploaded to different regions, but with the same hash. But yes, it does not make sense to set the output each time; once should be sufficient.

src/pipeline.ts
kaizen3031593: talked about how this would fail the use case where we upload `cdk.out` directly to github and skip `cdk synth`.

rix0rrr (approved):

PR #98, opened 2022-02-09
feat: diff protection

kaizen3031593:
Adds diff protection between local `cdk synth` and the `cdk synth` performed via GitHub Actions. To achieve this, I've added a known environment variable, `SYNTH` that is set to `'github'` during the workflow. If this environment variable is set, then we check to make sure that the `.yml` file that was committed is equal to the `.yml` file that gets generated.

TODO: add an escape hatch for users who want to disable this feature.


src/pipeline.ts
```diff
@@ -194,6 +194,19 @@ export class GitHubWorkflow extends PipelineBase {
     // eslint-disable-next-line no-console
     console.error(`writing ${this.workflowPath}`);
     mkdirSync(path.dirname(this.workflowPath), { recursive: true });
+
+    if (process.env.SYNTH === 'github') {
```
rix0rrr, line 198: I'm 99% sure there are just environment variables that will already be set for you to indicate that you are running on GitHub Actions. Why not check those?

src/pipeline.ts
```diff
@@ -194,6 +194,19 @@ export class GitHubWorkflow extends PipelineBase {
...
+      // check if workflow file has changed
+      if (!existsSync(this.workflowPath)) {
+        throw new Error('The committed workflow file differs from the synthesized workflow file. Did you forget to commit?');
```
rix0rrr, line 201: You can get rid of the duplicated error message. Either that, or make this one more accurate.

rix0rrr: I'm debating whether "Did you forget..." sounds accusatory 😅.

How about "please commit the updated workflow file (${relative path here}) when you change your pipeline definition."

rix0rrr (changes requested):

test/github.test.ts
```diff
@@ -103,3 +103,42 @@ describe('workflow path', () => {
...
+      // simulate GitHub environment with the same deploy.yml
+      wrapEnv('GITHUB_ACTIONS', 'true', () => app.synth());
+    });
+  });
+});
+
+function wrapEnv(variable: string, value: string, cb: () => void) {
+  const original = process.env[variable];
+  process.env[variable] = value;
+  cb();
```
rix0rrr, line 142: Needs a `try/finally`

rix0rrr (changes requested):

rix0rrr (approved):

PR #121, opened 2022-03-07
chore(readme): update to make more beginner friendly

kaizen3031593:


README.md
```diff
@@ -74,7 +75,32 @@ documentation for more details.
   Environment
   Bootstrapping](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.pipelines-readme.html#cdk-environment-bootstrapping)
   for details.
-  
+
+## Developer Workflow
```
rix0rrr, line 79: How about a section of "Initial setup", where you recommend they deploy the OIDC provider?

README.md
```diff
@@ -305,11 +333,28 @@ workflow in action. Make sure your GitHub repository has `AWS_ACCESS_KEY_ID` and
 `AWS_SECRET_ACCESS_KEY` secrets that can access the same account that you
 synthesized against.
 
-## Not supported yet
+> Here, we do want to commit `cdk.out`. This is not a recommended use-case; it is because
```
rix0rrr, line 336: What is this trying to say? I would just steer away from recommending this.

README.md
```diff
@@ -74,7 +75,32 @@ documentation for more details.
...
+Assuming you have your CDK app checked out on your local machine, here are the suggested steps
+to develop your GitHub Workflow.
+
+* Set up AWS Credentials your local environment. See [AWS Credentials](#aws-credentials) for the two
+  ways you may set up your credentials.
+
+* When you've updated your pipeline and are ready to deploy, run `cdk synth`. `cdk synth` will generate
+  a cloud assembly under the `cdk.out` folder and also the workflow file in `.github/workflows/deploy.yml`.
+  You should commit your code changes to GitHub along with `deploy.yml`, but you do not need to commit
+  the contents in `cdk.out`.
```
rix0rrr, line 90: It sounds more like you're explaining "how it works" than "what a user should do".

It would put those in different sections. "What it looks like to you" is more important for users than "how it works."

image

rix0rrr (changes requested):

README.md
kaizen3031593: I will split the contents apart, though

README.md
kaizen3031593: In the 'tutorial', the user is running the example app from a different repository, and the example app has these options set. I imagine many people start out by trying this and/or copying the example app entirely, which is bad, so I want to build awareness on it. I'm trying to say *don't do this* here, so I'll adjust it to be a bit more forceful.

README.md
kaizen3031593: i couldn't think of a reasonable way to explain how it works, so I omitted it for now. perhaps you are right and it is unnecessary.

rix0rrr (approved):

PR #135, opened 2022-03-24
feat: Add option to pass in the runs-on attribute to enable self-hosted runner

FahadMustafa:
Adds the option to pass in a `runsOn` value to override the default `ubuntu-latest` value.
This is mainly done to enable `self-hosted` runners


src/pipeline.ts
```diff
@@ -101,6 +100,14 @@ export interface GitHubWorkflowProps extends PipelineBaseProps {
...
    */
   readonly dockerCredentials?: DockerCredential[];
+
+  /**
+   * The type of machine to run the job on. The machine can be either a
+   * GitHub-hosted runner or a self-hosted runner.
+   *
+   * @default "ubuntu-latest"
+   */
+  readonly runsOn?: string;
```
kaizen3031593, line 110: lets make this type an enum-like class `Runner` that looks something like this. I'm modeling this off of the aws-cdk lambda [`Runtime`](https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/aws-lambda/lib/runtime.ts) that we have. I haven't fully thought this through but I'd like it to be modeled something like this:

```ts
export class RuntimeType {
  public static readonly GITHUB_HOSTED = 'github-hosted';
  public static readonly SELF_HOSTED = 'self-hosted';
}

export class Runtime {
  public static readonly UBUNTU_LATEST =  new Runner(['ubuntu-latest'], RuntimeType.GITHUB_HOSTED);

  public static readonly WINDOWS_LATEST = new Runner(['windows-latest'], RuntimeType.GITHUB_HOSTED);

  public static readonly MACOS_LATEST = new Runner(['macos-latest'], RuntimeType.GITHUB_HOSTED);

  public static selfHosted(labels: string[]): Runner {
    return new Runtime(['self-hosted', ...labels]);
  } 

  public constructor(public readonly runsOn: string) {}
}
```

One thing we'll have to figure out is how to get `runsOn: ubuntu-latest` most of the time, but if you want a self-hosted runtime you can have a [list](https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/aws-lambda/lib/runtime.ts) of labels. Again, I haven't fully thought this through, so feel free to make some changes to the model I'm suggesting.

kaizen3031593 (changes requested):
Hi @FahadMustafa!

First of all, thanks so much for opening this pull request! This is definitely something we want. In addition to building out the runtime type, please also add a section in the readme detailing what you added.

src/pipeline.ts
FahadMustafa: Thanks for the input @kaizen3031593. I've updated the PR.

src/workflows-model.ts
```diff
@@ -192,6 +191,49 @@ export interface JobDefaults {
...
+
+  public static selfHosted(labels: string[]): Runner {
+    return new Runner('self-hosted', RunnerType.SELF_HOSTED, labels);
+  }
+
+  public readonly runsOn: string;
+  public readonly runnerType: RunnerType;
+  public readonly labels: string[];
+
+  constructor(runsOn: string, runnerType: RunnerType, labels?: string[]) {
```
kaizen3031593, line 230: Let's make the constructor private and get rid of `RunnerType`

src/workflows-model.ts
```diff
@@ -192,6 +191,49 @@ export interface JobDefaults {
...
+    this.runsOn = runsOn;
+    this.runnerType = runnerType;
+    this.labels = labels ?? [];
+  }
+}
```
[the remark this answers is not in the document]
kaizen3031593: If this does not work (maybe there's some JSII error I'm forgetting about for `runsOn`) then lets just always return this.labels.

kaizen3031593 (changes requested):
Hi @FahadMustafa! Thanks again for opening this PR! For the next review, please do the following:

- decouple `publishAssetsRegion` from this PR and focus on the `runsOn` property
- remove all the formatting changes. they are cluttering the PR diff and make it hard to see what really has changed, and I find them unnecessary. if you really want to change the formatting of the code, please open a new PR for that alone and we can discuss there.
- take a look at my comments on how we should model the `Runner` class.

src/pipeline.ts
```diff
@@ -101,6 +114,20 @@ export interface GitHubWorkflowProps extends PipelineBaseProps {
...
+   * GitHub-hosted runner or a self-hosted runner.
+   * @default Runner.UBUNTU_LATEST
+   */
+  readonly runner?: Runner;
+
+  /**
+   * Optional override for the region used in Publish Assets job
+   * @default "us-west-2"
+   */
+  readonly publishAssetsRegion?: string;
```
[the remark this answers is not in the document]
FahadMustafa: Done. Will raise another PR after this is merged.

src/pipeline.ts
```diff
@@ -262,32 +323,45 @@ export class GitHubWorkflow extends PipelineBase {
...
-        throw new Error(`jobForNode: did not expect to get group nodes: ${node.data?.type}`);
+        throw new Error(
+          `jobForNode: did not expect to get group nodes: ${node.data?.type}`,
+        );
 
       case 'self-update':
         throw new Error('GitHub Workflows does not support self mutation');
 
       case 'publish-assets':
+        console.log(options);
```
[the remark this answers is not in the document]
FahadMustafa: Removed altogether

src/workflows-model.ts
```diff
@@ -10,7 +10,7 @@ export interface Job {
    *
    * @example "ubuntu-latest"
    */
-  readonly runsOn: string;
+  readonly runsOn: unknown;
```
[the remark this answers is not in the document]
FahadMustafa: this is not a `string []`

src/workflows-model.ts
FahadMustafa: Done

src/workflows-model.ts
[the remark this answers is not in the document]
FahadMustafa: Updated. Had to return `this.labels`.

kaizen3031593:
@FahadMustafa please also revert the formatting changes that are cluttering the PR. I do not find them necessary and if you disagree, I am happy to hold that discussion on a separate PR. Let's keep this PR to exactly what is necessary for the runs-on attribute

FahadMustafa:
> @FahadMustafa please also revert the formatting changes that are cluttering the PR. I do not find them necessary and if you disagree, I am happy to hold that discussion on a separate PR. Let's keep this PR to exactly what is necessary for the runs-on attribute

@kaizen3031593 apologies for the formatting issues. I don't disagree with any of the current formatting. vscode was auto-formatting on save.

src/workflows-model.ts
```diff
@@ -192,6 +192,36 @@ export interface JobDefaults {
...
+ * In case of self-hosted, a list of labels can be supplied.
+ */
+export class Runner {
+  public static readonly UBUNTU_LATEST = new Runner(['ubuntu-latest']);
+
+  public static readonly WINDOWS_LATEST = new Runner(['windows-latest']);
+
+  public static readonly MACOS_LATEST = new Runner(['macos-latest']);
+
+  public static selfHosted(labels: string[]): Runner {
```
kaizen3031593, line 206: All these public APIs need docs. Specifically for `selfHosted`, please mention that you can supply `self-hosted` as a label, but you do not have to.

src/pipeline.ts
```diff
@@ -332,9 +401,11 @@ export class GitHubWorkflow extends PipelineBase {
         needs: this.renderDependencies(node),
         permissions: {
           contents: github.JobPermission.READ,
-          idToken: this.useGitHubActionRole ? github.JobPermission.WRITE : github.JobPermission.NONE,
+          idToken: this.useGitHubActionRole
+            ? github.JobPermission.WRITE
+            : github.JobPermission.NONE,
         },
-        runsOn: RUNS_ON,
+        runsOn: this.runsOn,
```
kaizen3031593, line 408:
```suggestion
        runsOn: this.Runner.runsOn,
```

src/pipeline.ts
```diff
@@ -122,6 +142,8 @@ export class GitHubWorkflow extends PipelineBase {
   private readonly postBuildSteps: github.JobStep[];
   private readonly jobOutputs: Record<string, github.JobStepOutput[]> = {};
   private readonly assetHashMap: Record<string, string> = {};
+  private readonly runner: Runner;
+  private readonly runsOn: string [];
```
kaizen3031593, line 146:
```suggestion

```

README.md
```diff
@@ -276,6 +281,54 @@ const pipeline = new GitHubWorkflow(app, 'Pipeline', {
 });
 ```
 
+## Runner Types
+
+You can choose to run the workflow in either a GitHub hosted or [self-hosted](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) runner.
+
+### GitHub Hosted Runner
+
+The default is `ubuntu-latest`. You can override this as shown below:
```
kaizen3031593, line 290:
```suggestion
The default is `Runner.UBUNTU_LATEST`. You can override this as shown below:
```

src/pipeline.ts
```diff
@@ -142,24 +164,34 @@ export class GitHubWorkflow extends PipelineBase {
     this.dockerCredentials = props.dockerCredentials ?? [];
 
     this.workflowPath = props.workflowPath ?? '.github/workflows/deploy.yml';
-    if (!this.workflowPath.endsWith('.yml') && !this.workflowPath.endsWith('.yaml')) {
+    if (
```
kaizen3031593, line 167: revert formatting

src/pipeline.ts
```diff
@@ -142,24 +164,34 @@ export class GitHubWorkflow extends PipelineBase {
...
+      !this.workflowPath.endsWith('.yml') &&
+      !this.workflowPath.endsWith('.yaml')
+    ) {
       throw new Error('workflow file is expected to be a yaml file');
     }
     if (!this.workflowPath.includes('.github/workflows/')) {
-      throw new Error('workflow files must be stored in the \'.github/workflows\' directory of your repository');
+      throw new Error(
```
kaizen3031593, line 174: revert formatting

src/pipeline.ts
```diff
@@ -142,24 +164,34 @@ export class GitHubWorkflow extends PipelineBase {
...
+
+    this.runner = props.runner ?? Runner.UBUNTU_LATEST;
+    this.runsOn = this.runner.runsOn;
   }
 
   protected doBuildPipeline() {
     const app = Stage.of(this);
     if (!app) {
-      throw new Error('The GitHub Workflow must be defined in the scope of an App');
+      throw new Error(
```
kaizen3031593, line 192: revert formatting

src/pipeline.ts
```diff
@@ -173,7 +205,9 @@ export class GitHubWorkflow extends PipelineBase {
 
     for (const stageNode of flatten(structure.graph.sortedChildren())) {
       if (!isGraph(stageNode)) {
-        throw new Error(`Top-level children must be graphs, got '${stageNode}'`);
+        throw new Error(
```
kaizen3031593, line 208: revert formatting

src/pipeline.ts
```diff
@@ -220,11 +254,20 @@ export class GitHubWorkflow extends PipelineBase {
 
     // GITHUB_WORKFLOW is set when GitHub Actions is running the workflow.
     // see: https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
-    const diffProtection = this.node.tryGetContext('cdk-pipelines-github:diffProtection') ?? true;
+    const diffProtection =
```
kaizen3031593, line 257: revert formatting

src/pipeline.ts
```diff
@@ -220,11 +254,20 @@ export class GitHubWorkflow extends PipelineBase {
...
+      this.node.tryGetContext('cdk-pipelines-github:diffProtection') ?? true;
     if (diffProtection && process.env.GITHUB_WORKFLOW === this.workflowName) {
       // check if workflow file has changed
-      if (!existsSync(this.workflowPath) || yaml !== readFileSync(this.workflowPath, 'utf8')) {
-        throw new Error(`Please commit the updated workflow file ${path.relative(__dirname, this.workflowPath)} when you change your pipeline definition.`);
+      if (
```
kaizen3031593, line 261: revert formatting

src/pipeline.ts
```diff
@@ -248,7 +291,9 @@ export class GitHubWorkflow extends PipelineBase {
   private renderJobOutputs(outputs: github.JobStepOutput[]) {
     const renderedOutputs: Record<string, string> = {};
     for (const output of outputs) {
-      renderedOutputs[output.outputName] = `\${{ steps.${output.stepId}.outputs.${output.outputName} }}`;
+      renderedOutputs[
```
kaizen3031593, line 294: revert formatting

src/pipeline.ts
```diff
@@ -262,7 +307,9 @@ export class GitHubWorkflow extends PipelineBase {
       case 'group':
       case 'stack-group':
       case undefined:
-        throw new Error(`jobForNode: did not expect to get group nodes: ${node.data?.type}`);
+        throw new Error(
```
kaizen3031593, line 310: revert formatting

src/pipeline.ts
```diff
@@ -274,20 +321,30 @@ export class GitHubWorkflow extends PipelineBase {
         throw new Error('"prepare" is not supported by GitHub Worflows');
 
       case 'execute':
-        return this.jobForDeploy(node, node.data.stack, node.data.captureOutputs);
+        return this.jobForDeploy(
```
kaizen3031593, line 324: revert formatting

src/pipeline.ts
```diff
@@ -274,20 +321,30 @@ export class GitHubWorkflow extends PipelineBase {
...
+        );
 
       case 'step':
         if (node.data.isBuildStep) {
           return this.jobForBuildStep(node, node.data.step);
         } else if (node.data.step instanceof ShellStep) {
           return this.jobForScriptStep(node, node.data.step);
         } else {
-          throw new Error(`unsupported step type: ${node.data.step.constructor.name}`);
+          throw new Error(
```
kaizen3031593, line 336: revert formatting

src/pipeline.ts
```diff
@@ -274,20 +321,30 @@ export class GitHubWorkflow extends PipelineBase {
...
+            `unsupported step type: ${node.data.step.constructor.name}`,
+          );
         }
     }
   }
 
-  private jobForAssetPublish(node: AGraphNode, assets: StackAsset[], options: Context): Job {
+  private jobForAssetPublish(
```
kaizen3031593, line 343: revert formatting

kaizen3031593 (changes requested):
This is almost there! There's still a lot of formatting changes I'd like you to revert in `src/pipeline.ts`. I tried commenting on a few but there are really a lot.

The last thing I want to think about is whether or not it is worth it to somehow turn
```yaml
runsOn:
  - ubuntu-latest
```
into 
```yaml
runsOn: ubuntu-latest
```
Ideally, we do not disrupt templates like this when they upgrade `cdk-pipelines-github` since its alarming when your template changes and your code does not change. In this case it's probably fine. But something to think about.

src/workflows-model.ts
```diff
@@ -10,7 +10,7 @@ export interface Job {
    *
    * @example "ubuntu-latest"
```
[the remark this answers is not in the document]
FahadMustafa: Updated

src/workflows-model.ts
FahadMustafa: Added comment to code and mentioned in README

src/pipeline.ts
FahadMustafa: updated

src/pipeline.ts
```diff
@@ -142,24 +164,34 @@ export class GitHubWorkflow extends PipelineBase {
...
     }
 
     this.workflowName = props.workflowName ?? 'deploy';
     this.workflowTriggers = props.workflowTriggers ?? {
       push: { branches: ['main'] },
       workflowDispatch: {},
     };
+
+    this.runner = props.runner ?? Runner.UBUNTU_LATEST;
+    this.runsOn = this.runner.runsOn;
```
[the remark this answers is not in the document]
FahadMustafa: removed

src/pipeline.ts
FahadMustafa: done

README.md
FahadMustafa: updated

FahadMustafa:
@kaizen3031593 Somehow I completely missed reverting the formatting on `pipeline.ts`. This is now done. Apologies!

FahadMustafa:
> This is almost there! There's still a lot of formatting changes I'd like you to revert in `src/pipeline.ts`. I tried commenting on a few but there are really a lot.
> 
> The last thing I want to think about is whether or not it is worth it to somehow turn
> 
> ```yaml
> runsOn:
>   - ubuntu-latest
> ```
> 
> into
> 
> ```yaml
> runsOn: ubuntu-latest
> ```
> 
> Ideally, we do not disrupt templates like this when they upgrade `cdk-pipelines-github` since its alarming when your template changes and your code does not change. In this case it's probably fine. But something to think about.

I've made the change to only generate `runsOn` as an array for self-hosted runners.

kaizen3031593 (changes requested):
This is near the finish line -- thanks for all your help so far!

Give me until early next week to try this out locally, to make sure the snapshots actually do deploy and succeed (it would help my confidence if you did so as well and report back)!

src/workflows-model.ts
```diff
@@ -8,9 +8,9 @@ export interface Job {
    * The type of machine to run the job on. The machine can be either a
    * GitHub-hosted runner or a self-hosted runner.
    *
-   * @example "ubuntu-latest"
+   * @example ["ubuntu-latest"]
    */
-  readonly runsOn: string;
+  readonly runsOn: string[] | string;
```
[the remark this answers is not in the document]
FahadMustafa: yes, agree. The alternative is to always render out an array. I feel this is an acceptable trade-off for better end-user experience.

FahadMustafa:
> This is near the finish line -- thanks for all your help so far!
> 
> Give me until early next week to try this out locally, to make sure the snapshots actually do deploy and succeed (it would help my confidence if you did so as well and report back)!

Well, we've actually used this to deploy a static website to S3 on self-hosted runners. We've referenced the source code directly to unblock ourselves. It's working well so far!

kaizen3031593 (approved):
Thanks for the contribution! This looks great.

PR #157, opened 2022-04-11
feat: override publish asset authentication region

FahadMustafa:
This PR introduces a new property called `publishAssetsAuthRegion`. This is to give consumers the option to override the default `us-west-2` region when assuming role to publish assets.


src/pipeline.ts
```diff
@@ -107,6 +107,12 @@ export interface GitHubWorkflowProps extends PipelineBaseProps {
    * @default Runner.UBUNTU_LATEST
    */
   readonly runner?: github.Runner;
+
+  /**
+   * Optional override for the region used when assuming role to publish assets
```
kaizen3031593, line 112:
```suggestion
   * Will assume the GitHubActionRole in this region when publishing assets.
   * This is NOT the region in which the assets are published.
   *
   * In most cases, you do not have to worry about this property, and can safely
   * ignore it.
   *
   * @default 'us-west-2'
```

kaizen3031593 (approved):

PR #172, opened 2022-04-22
fix: synth step should have OIDC permissions to unblock custom steps that may want to use OIDC authentication

### LINKED ISSUE #174: Allow build step to Authenticate via OIDC and Assume Role for Lookups during Synthesis
**Issue Description** - During synthesis and build job execution time, we require to do vpc lookup as under 
const sharedVpc = Vpc.fromLookup(this, `SharedVpc-${name}`, {
          vpcName: sharedVpcName,
        });

For that I need to do authentication into my target AWS environment via OIDC, I can very well add a pre-build steps to Authenticate via OIDC and Assume role using awscredentials step action in my cdk code, but as the permissions set at job level don't provide any id token write access, and neither I am able to extend job level structure in my cdk code so it will help if we can add this id-token write permission in the jobForBuildStep function

**Error** - [Error at /Dev/PrefillApiStack/prefill-api] Could not assume role in target account using current credentials (which are for account ****) User: arn:aws:sts::<accountid>:assumed-role/Dev-github-runners-GitHubRunnerRole****/I-******** is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::<accountidtarget>:role/***-lookup-role-**-ap-southeast-2 . Please make sure that this role exists in the account. If it doesn't exist, (re)-bootstrap the environment with the right '--trust', using the latest version of the CDK CLI.

**Analysis done so far and potential fix** - There is a need to Authenticate via OIDC and Assume role during Synthesis in order to perform cdk look ups. The current permissions at job level for Synthesis /Build is only token read and it needs to be have id-token write permission for authenticating via OIDC role. Hence added the fix/change in pipeline.ts in jobForBuildStep function to add permission idToken: this.useGitHubActionRole ? github.JobPermission.WRITE : github.JobPermission.NONE

Once this is done, I added a pre build step to my synth as under and it fixed the issue.

```ts
private stepToAssumeLookUpRole(env:Environment): github.JobStep[] {
    return [awsCredentialStep('Authenticate Via OIDC Role', {
      region: env.region ?? 'ap-southeast-2',
      gitHubActionRoleArn: this.gitHubActionRoleArnValue,
    }), awsCredentialStep('Assume CDK Lookup Role', {
      region:  env.region ?? 'ap-southeast-2',
      accessKeyId: '${{ env.AWS_ACCESS_KEY_ID }}',
      secretAccessKey: '${{ env.AWS_SECRET_ACCESS_KEY }}',
      sessionToken: '${{ env.AWS_SESSION_TOKEN }}',
      roleToAssume: 'arn:aws:iam::'+env.account+':role/***-lookup-role-'+env.account+'-ap-southeast-2',
    })
    ];
  }
```

awshilpa:
Fixes #174.

There is a need to Authenticate via OIDC and Assume role during Synthesis in order to perform cdk look ups. The current permissions at job level for Synthesis /Build is only token read and it needs to be have id-token write permission for authenticating via OIDC role. Hence added the fix/change in pipeline.ts in jobForBuildStep function to add permission idToken: this.useGitHubActionRole ? github.JobPermission.WRITE : github.JobPermission.NONE

Please review.


src/pipeline.ts
```diff
@@ -465,6 +465,7 @@ export class GitHubWorkflow extends PipelineBase {
         name: 'Synthesize',
         permissions: {
           contents: github.JobPermission.READ,
+          idToken: this.useGitHubActionRole ? github.JobPermission.WRITE : github.JobPermission.NONE,
```
kaizen3031593, line 468: This addition on its own will not do anything, because the steps in this job do not configure aws credentials.

It's actually non-trivial to add aws credentials here because the API `this.stepsToConfigureAws()` requires a region and account. We don't really have that in `jobForBuildStep`.

kaizen3031593 (commented):
Hi @awshilpa! Thanks for opening this PR and bringing this to my attention. I assume that your use case is trying to access `HostedZone.fromLookup()` or a similar API that uses the stack environment to look this up during synth. I agree that this is currently blocked (sorry!) but I don't think your solution is enough. 

Can you open an issue explaining your use case and the error you're getting? Then we can discuss potential solutions there. I definitely want to unblock you but I'm not sure if we have a solution yet!

src/pipeline.ts
awshilpa: Hi, yes this change alone is not enough, but we could make it work by adding aws credentials step as a prebuild step to the synth in while creating GitHubAction Workflow in our pipeline stack.

kaizen3031593:
@awshilpa build needs to be updated with snapshot test changes; I can't do it for you because you are working from your `main` branch.

awshilpa:
Thanks @kaizencc  for your feedback ! There is a conflict on yarn lock file regarding eslint version, we should be able to use updated one? Additionally update the test snapshots, please let me know if that's fine to merge

kaizencc:
Closing in favor of #182, since I cannot make changes to this PR as it is cut from `main`

PR #178, opened 2022-04-25
feat: add new api to unlock github-specific stage-level properties

kaizencc:
Fixes #162.

Introduces a new API as follows:

```ts
import { App } from 'aws-cdk-lib';
import { ShellStep } from 'aws-cdk-lib/pipelines';
import { GitHubWorkflow } from 'cdk-pipelines-github';

const app = new App();

const pipeline = new GitHubWorkflow(app, 'Pipeline', {
  synth: new ShellStep('Build', {
    commands: [
      'yarn install',
      'yarn build',
    ],
  }),
  gitHubActionRoleArn: 'arn:aws:iam::<account-id>:role/GitHubActionRole',
});

pipeline.addStageWithGitHubOptions(new MyStage(this, 'Beta', {
  env: BETA_ENV,
  gitHubEnvironment: 'beta',
}));
pipeline.addStageWithGitHubOptions(new MyStage(this, 'Prod', {
  env: PROD_ENV,
  gitHubEnvironment: 'prod',
}));

app.synth();
```

Unlocks GitHub Environments at a stage-level, as well as a blueprint for any future stage-level github-specific properties we want to add.

PR #186, opened 2022-04-29
fix: Add the environment attribute to workflow yml file conditionally

julianiag:
Based on discussion https://github.com/cdklabs/cdk-pipelines-github/pull/178#discussion_r861411138

Removes `environment: null` from workflows that did not specify the environment.

### DIFF
--- a/src/pipeline.ts
+++ b/src/pipeline.ts
@@ -461,9 +461,9 @@ export class GitHubWorkflow extends PipelineBase {
           contents: github.JobPermission.READ,
           idToken: this.useGitHubActionRole ? github.JobPermission.WRITE : github.JobPermission.NONE,
         },
-        ...(this.stackEnvs[stack.stackArtifactId] && {
+        ...(this.stackEnvs[stack.stackArtifactId] ? {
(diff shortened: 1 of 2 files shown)


src/pipeline.ts
```diff
@@ -461,7 +461,9 @@ export class GitHubWorkflow extends PipelineBase {
           contents: github.JobPermission.READ,
           idToken: this.useGitHubActionRole ? github.JobPermission.WRITE : github.JobPermission.NONE,
         },
-        environment: this.stackEnvs[stack.stackArtifactId],
+        ...(this.stackEnvs[stack.stackArtifactId] && {
+          environment: this.stackEnvs[stack.stackArtifactId],
+        }),
```
kaizencc, line 466:
```suggestion
        ...(this.stackEnvs[stack.stackArtifactId] ? {
          environment: this.stackEnvs[stack.stackArtifactId],
        } : {}),
```

kaizencc (approved):
Minor nit: the result should be the same but it's better to be explicit about what happens when `this.stackEnvs[stack.stackArtifactId]` doesn't exist.

Thanks for contributing this!

PR #187, opened 2022-04-29
feat: add stackCapabilities to addGitHubStageOptions

### LINKED ISSUE #179: Cannot deploy applications w/named IAM resources
I attempted to deploy a stack that contains a named IAM resource and the stack creation failed because the `Deploy` step (which uses the `aws-actions/aws-cloudformation-github-deploy` action) in the pipeline did not have the `capabilities` property set and the default value lacks the `CAPABILITY_NAMED_IAM` capability.

We should be able to set these values explicitly to ensure that stacks can be created even if they have named IAM resources.

e.g. 
```
import cdk_pipelines_github as GithubPipelines

app = cdk.App()

pipeline = GithubPipelines.GitHubWorkflow(app, ...)

pipeline.add_stage(
    stage=MyAppStage(...),
    deployment_capabilities=[ # <--- NEW
      "CAPABILITY_IAM",
      "CAPABILITY_NAMED_IAM",
      "CAPABILITY_AUTO_EXPAND"
    ]
)
```

kaizencc:
Fixes #179.

Unlocks capabilities property to `addGitHubStageOptions`. 

Not advertising in the README because I don't think it's that common of a use case, but it is now unblocked.


test/__snapshots__/stage-options.test.ts.snap
```diff
@@ -0,0 +1,308 @@
...
+    steps:
+      - name: Authenticate Via GitHub Secrets
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-region: us-east-1
+          role-duration-seconds: 1800
+          role-skip-session-tagging: true
+          aws-access-key-id: \${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: \${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-session-token: \${{ secrets.undefined }}
```
Chriscbr, line 80: Bug?

src/pipeline.ts
```diff
@@ -292,7 +285,7 @@ export class GitHubWorkflow extends PipelineBase {
     }
 
     // eslint-disable-next-line no-console
-    console.log(`writing ${this.workflowPath}`);
+    // console.log(`writing ${this.workflowPath}`);
```
Chriscbr, line 288: Should this be removed?

Chriscbr (approved):
LGTM, just two small nits.

test/__snapshots__/stage-options.test.ts.snap
kaizencc: This is an ugly thing that I will remove in a separate PR, it's not specific to this one.

PR #280, opened 2022-07-27
Publish assets file in the assembly folder

### LINKED ISSUE #279: Script for publishing assets should be stage specific
I have a pipeline that consists of two stages (staging and prod) deploying the same stack to different accounts depending on the branch. This correctly generates two independent workflows but the problem comes with the `publish-Assets-FileAsset1-step.sh`. As the asset ID is the same in both stages the file ends up with whatever the last stage wrote there, thus pointing to the wrong account/region.

svyatogor:
Fixes #279


kaizencc:
Apologies for not commenting on this PR before, but I can't take a look at it without testing.

kaizencc:
Closing this as abandoned.

PR #304, opened 2022-08-17
feat: github action job step as pre/post deploy steps

### LINKED ISSUE #222: add GitHubActionStep that extends Step
### Use case
We would like to add a e2e test after deploy step, this can be achieved by adding `option` with `ShellStep` . The problem for this case is that there is no way to use the checkout action or any other action. So I am asking if we could introduce preDeploySteps and postDeploySteps like what we already had `preBuildSteps` and `postBuildSteps`?

Please let me know if you have any better idea to achieve this?

julianiag:
This is an attempt to resolve #222

Fixes #222


src/steps/GithubActionStep.ts
```diff
@@ -0,0 +1,25 @@
+import { Step } from 'aws-cdk-lib/pipelines';
```
kaizencc, line 1: can we rename this file `github-action-step.ts`

src/steps/GithubActionStep.ts
```diff
@@ -0,0 +1,25 @@
...
+import { JobStep } from '../workflows-model';
+
+interface GithubActionStepProps {
```
kaizencc, line 4:
```suggestion
interface GitHubActionStepProps {
```

src/steps/GithubActionStep.ts
```diff
@@ -0,0 +1,25 @@
...
+   * Environment variables to set.
+   */
+  env?: Record<string, string>;
+}
+
+export class GithubActionStep extends Step {
+  readonly env: Record<string, string>;
+  readonly jobStep: JobStep;
+
+  constructor(id: string, { env, jobStep }: GithubActionStepProps) {
```
kaizencc, line 20: in case we plan on adding more `GitHubActionStepProps` in the future, lets do

```ts
constructor(id: string, props: GitHubActionStepProps) {
  super(id);
  this.jobsStep = props.jobStep;
  this.env = props.env ?? {};
}
```

test/stage-options.test.ts
```diff
@@ -2,6 +2,7 @@ import { readFileSync } from 'fs';
 import { Stack, Stage } from 'aws-cdk-lib';
 import { ShellStep } from 'aws-cdk-lib/pipelines';
 import { GitHubWorkflow, StackCapabilities } from '../src';
+import { GithubActionStep } from '../src/steps/GithubActionStep';
```
kaizencc, line 5: We need to expose this in `index.ts` and then I think we can just import from `../src`. Otherwise, `GitHubActionStep` won't actually be exported.

test/stage-options.test.ts
```diff
@@ -196,4 +197,50 @@ describe('job settings', () => {
...
+              'secrets': 'secrets',
+            },
+          },
+        })],
+      });
+
+      app.synth();
+
+      expect(readFileSync(pipeline.workflowPath, 'utf-8')).toContain('my-pre-deploy-action\@1\.0\.0');
+      expect(readFileSync(pipeline.workflowPath, 'utf-8')).toContain('my-post-deploy-action\@1\.0\.0');
```
kaizencc, line 243: I know I've said otherwise before, but I think this change deserves a snapshot test in addition (can be the same test).

test/stage-options.test.ts
```diff
@@ -196,4 +197,50 @@ describe('job settings', () => {
...
+              'app-id': 4321,
+              'secrets': 'secrets',
+            },
+          },
+        })],
+      });
+
+      app.synth();
+
+      expect(readFileSync(pipeline.workflowPath, 'utf-8')).toContain('my-pre-deploy-action\@1\.0\.0');
```
kaizencc, line 242: we need to somehow also assert that the pre-deploy-action happens in the right place (pre deployment) and similarly for post-deploy-action.

kaizencc (changes requested):
Hi @julianiag! Sorry for missing this PR for a month, I think it looks excellent. In addition to the minor comments below, we need to add a section in the readme that explains this feature. Thanks for submitting this contribution!

src/steps/GithubActionStep.ts
julianiag: updated.

src/steps/GithubActionStep.ts
julianiag: I wonder there is no difference in your suggested change? ❓

src/steps/GithubActionStep.ts
julianiag: updated

test/stage-options.test.ts
julianiag: updated

test/stage-options.test.ts
julianiag: added snapshot test

test/stage-options.test.ts
julianiag: I feel the snapshot test is enough to guarantee the position?

src/steps/GithubActionStep.ts
kaizencc: capital H :)

kaizencc (approved):
Yay! Thanks @julianiag!

PR #307, opened 2022-08-19
feat: Add support to provide GHA role per stage

### LINKED ISSUE #302: Use separate GHA roles per stage
I would like to use a slightly different authentication pattern where each stage uses its own GHA role. The current configuration appears to only support a single role for the whole pipeline, which is then used to assume the CDK role for upload/deployment.

Would it be possible to modify the stage configuration `pipeline.addStageWithGitHubOptions()` to support a separate role per stage? Would you be open to accepting a PR which does this?

Background:

The reason for this request is that with a single role, it seems like we would need to either:

1. Have one stage account designated as the 'pipeline' account, which contains the GHA role and then have the other accounts setup with `cdk bootstrap` to trust it. **OR**
2. Have a completely separate `pipeline` account to contain the GHA role which is trusted by the stage accounts.

The above change, to support a role per stage, would make the configuration closer to the cdk pattern, where we 'bootstrap' the accounts first, to create the roles, with no need for an additional account, or mixing 'pipeline' and 'stage' accounts.

danieljamesscott:
Fixes #302.

### DIFF
--- a/src/pipeline.ts
+++ b/src/pipeline.ts
@@ -218,7 +218,7 @@ export class GitHubWorkflow extends PipelineBase {
     this.addStackProps(stacks, 'environment', options?.gitHubEnvironment);
     this.addStackProps(stacks, 'capabilities', options?.stackCapabilities);
     this.addStackProps(stacks, 'settings', options?.jobSettings);
-    this.addStackProps(stacks, 'role', options?.role);
+    this.addStackProps(stacks, 'role', options?.gitHubActionsRoleArn);
 
     return stageDeployment;
   }
--- a/src/pipeline.ts
+++ b/src/pipeline.ts
@@ -218,6 +218,7 @@ export class GitHubWorkflow extends PipelineBase {
     this.addStackProps(stacks, 'environment', options?.gitHubEnvironment);
     this.addStackProps(stacks, 'capabilities', options?.stackCapabilities);
     this.addStackProps(stacks, 'settings', options?.jobSettings);
+    this.addStackProps(stacks, 'role', options?.role);
 
     return stageDeployment;
   }
@@ -370,6 +371,7 @@ export class GitHubWorkflow extends PipelineBase {
     const cdkoutDir = options.assemblyDir;
     const jobId = node.uniqueId;
     const assetId = assets[0].assetId;
+    const stageRoleArn = this.ghaStageRoleArn();
 
     // check if asset is docker asset and if we have docker credentials
     const dockerLoginSteps: github.JobStep[] = [];
@@ -419,7 +421,7 @@ export class GitHubWorkflow extends PipelineBase {
             name: 'Install',
             run: `npm install --no-save cdk-assets${installSuffix}`,
           },
-          ...this.stepsToConfigureAws(this.useGitHubActionRole, { region: this.publishAssetsAuthRegion }),
+          ...this.stepsToConfigureAws(this.useGitHubActionRole, { region: this.publishAssetsAuthRegion, stageRoleArn: stageRoleArn }),
           ...dockerLoginSteps,
           publishStep,
         ],
@@ -470,6 +472,8 @@ export class GitHubWorkflow extends PipelineBase {
     }
     const assumeRoleArn = stack.assumeRoleArn ? resolve(stack.assumeRoleArn) : undefined;
 
+    const stageRoleArn = this.ghaStageRoleArn();
+
     return {
       id: node.uniqueId,
       definition: {
@@ -486,7 +490,7 @@ export class GitHubWorkflow extends PipelineBase {
         needs: this.renderDependencies(node),
         runsOn: this.runner.runsOn,
         steps: [
-          ...this.stepsToConfigureAws(this.useGitHubActionRole, { region, assumeRoleArn }),
+          ...this.stepsToConfigureAws(this.useGitHubActionRole, { region, assumeRoleArn, stageRoleArn }),
           {
             id: 'Deploy',
             uses: 'aws-actions/aws-cloudformation-github-deploy@v1',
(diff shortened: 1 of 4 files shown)


danieljamesscott:
I fixed the commit message. I'm not sure how to re-trigger the check?

test/stage-options.test.ts
```diff
@@ -197,3 +197,64 @@ describe('job settings', () => {
     });
   });
 });
+
+describe('role settings', () => {
+  test('can specify gha role', () => {
```
kaizencc, line 202: this test isn't actually testing any features implemented in this PR, is it?

src/stage-options.ts
```diff
@@ -37,6 +37,13 @@ export interface AddGitHubStageOptions extends AddStageOpts {
    * Currently the only valid setting is 'if'.
    */
   readonly jobSettings?: JobSettings;
+
+  /**
```
kaizencc, line 41: Can we be a tad more descriptive here.

src/pipeline.ts
```diff
@@ -639,7 +643,16 @@ export class GitHubWorkflow extends PipelineBase {
     };
   }
 
-  private stepsToConfigureAws(openId: boolean, { region, assumeRoleArn }: { region: string; assumeRoleArn?: string }): github.JobStep[] {
+  private ghaStageRoleArn(): string {
+    const stageRoleArn = Object.entries(this.stackProperties)[0][1].role;
```
kaizencc, line 647: this is excessively brittle... `stackProperties` should be accessed by `stackArtifactId`.

test/stage-options.test.ts
danieljamesscott: No. Sorry. This was meant to be a snapshot test to confirm the GHA pipeline role. I'll improve the tests

src/pipeline.ts
danieljamesscott: Sorry. I really struggled with this. I was doing as you suggested in `jobForDeploy()`:

    const gitHubActionsStageRoleArn = this.stackProperties[stack.stackArtifactId]?.role

But in `jobForAssetPublish()`, I didn't have access to `stack`, because it didn't seem available to pass in here: https://github.com/cdklabs/cdk-pipelines-github/blob/main/src/pipeline.ts#L345

Any ideas for how I can solve this?

test/stage-options.test.ts
```diff
@@ -197,3 +197,64 @@ describe('job settings', () => {
...
+      pipeline.addStageWithGitHubOptions(stage, {
+        jobSettings: {
+          if: 'github.repository == \'github/repo\'',
+        },
+        role: 'my-stage-role',
+      });
+
+      app.synth();
+
+      expect(readFileSync(pipeline.workflowPath, 'utf-8')).toMatchSnapshot();
```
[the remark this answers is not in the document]
danieljamesscott: Sure. I couldn't see a great way to so this. Should I parse the yaml from the generated file and then run some asserts? I could do some string comparisons, but that seems brittle, especially if I'm going to test stages with and without roles.

It would be great if we could get access to the generated data structure, before it's written to a file. That would help with escape hatches, as well as allowing me to write better tests, without using the filesystem at all.

danieljamesscott:
> Hi @danieljamesscott! Thanks for opening this PR! I'm not opposed to adding this feature but I think there are some things to think through before we can merge. 

Great. Thanks for the reply.

I've addressed a few issues, but have a few questions around others. If you can let me know which way you'd prefer things to work, I'll make the changes. I may be missing something, but the only testing strategy I could see was to ensure that the entire snapshot matches, or that a specific substring is present in the generated file.

> I'm most worried about the behavior when `gitHubActionRoleArn` is _not_ supplied. Today that means we will search for GitHub Secrets to authenticate. I don't see that behavior changed in the PR so I can only assume it will still do that. How will we be able to guess whether the user means to send their roles per stage instead of using github secrets?

I was following the same logic, but with an additional "stage role arn". So it should "fall back" if not supplied. e.g.
1. Use stage role if supplied, if not:
2. Use pipeline role if supplied, if not:
3. Use secrets

test/stage-options.test.ts
```diff
@@ -197,3 +197,64 @@ describe('job settings', () => {
...
+        },
+      });
+
+      app.synth();
+
+      expect(readFileSync(pipeline.workflowPath, 'utf-8')).toContain('if: github.repository == \'github/repo\'\n');
+    });
+  });
+
+  test('can specify role override settings at stage level', () => {
```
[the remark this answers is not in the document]
danieljamesscott: Yep. I will write a test for this. I'm struggling to figure out how best to do so though.

src/pipeline.ts
danieljamesscott: I've looked into this quite a bit more, and I can't figure out a nice way to work out which `Stage` corresponds to which "asset publish" step. Do you have any ideas here?

danieljamesscott: I've tried lots of different things here, but still can't figure out how to link the `publish` step to the correct stage in `jobForAssetPublish`. So I can't figure out which role should be used for the publish step.

kaizencc: Hi @danieljamesscott. This could be a problem. You're right, `jobForAssetPublish` doesn't have a concept of `Stage`, because it is stage-agnostic. What's happening is that the assets for all stages are combined into the asset publishing step, which is really supposed to be abstracted from the user. That's the benefit of having a single "pipeline" role that has the correct permissions just like in aws-cdk-lib/pipelines.

Having thought this through a bit, I'm a little less groovy of the idea of natively supporting a role per stage. This looks like it will introduce more problems than it solves. If we did support this, we'd need to introduce a new property like `assetPublishingRoleArn` and then we're not really solving the problem you put forth -- somewhere one of your AWS accounts is going to have to own the role for asset publishing.

Ideally, we introduce some sort of escape hatching mechanism that allows users to go off the beaten path and do what they want. Let's take this conversation back to the issue and see if we can come up with a design that makes sense.

kaizencc:
I _think_ this looks stale :). Given that its from 1.5 years ago. Would basically have to start from scratch for this, and I'm not sure we need it.

PR #340, opened 2022-09-23
feat: apply json patches to workflow file

### LINKED ISSUE #293: allow for escape hatching
This library is modeled after CDK Pipelines, and is meant to be feature complete compared to that library. There is one difference however -- CDK Pipelines is a CDK construct that is compiled down to a CloudFormation template, which allows for the escape-hatching mechanism that unlocks any and all out-of-bound changes to the underlying CloudFormation template. As of right now, CDK Pipelines GitHub does not have this feature (since it compiles to a GitHub Workflow) and is thus less customizable than its counterpart.

CDK Pipelines GitHub needs a way of escape hatching to the GitHub Workflow.

kaizencc:
Introduces Escape Hatches:

You can override the `deploy.yml` workflow file post-synthesis however you like.

```ts
import { App } from 'aws-cdk-lib';
import { ShellStep } from 'aws-cdk-lib/pipelines';
import { GitHubWorkflow, JsonPatch } from 'cdk-pipelines-github';

const app = new App();

    commands: [
      'yarn install',
      'yarn build',
    ],
  }),
});

const deployWorkflow = pipeline.workflowFile;
// add `on: workflow_call: {}` to deploy.yml
deployWorkflow.patch(JsonPatch.add('/on/workflow_call', {}));
// remove `on: workflow_dispatch` from deploy.yml
deployWorkflow.patch(JsonPatch.remove('/on/workflow_dispatch'));
```

This can be used to allow any feature we do not currently support, as it gives full access to the underlying file. Of course, with great power comes great responsibility...

Closes #293


README.md
```diff
@@ -340,6 +341,33 @@ const pipeline = new GitHubWorkflow(app, 'Pipeline', {
...
+    commands: [
+      'yarn install',
+      'yarn build',
+    ],
+  }),
+});
+
+const deployWorkflow = pipeline.workflowFile;
+// add `on: workflow_call: {}` to deploy.yml
+deployWorkflow.patch(JsonPatch.add('/on/workflow_call', '{}'));
```
kaizencc, line 366:
```suggestion
deployWorkflow.patch(JsonPatch.add('/on/workflow_call', {}));
```

rix0rrr (approved):

Issue #349, opened 2022-09-29
Ability to set a top-of-file comment on deploy.yaml


giseburt:
To aid on-boarding and also to be able to describe what's being deployed.

For example, the following as the first few lines of the `deploy.yml` file:

```yaml
# This file is auto generated. Please see the ReadMe.md for instructions.
# DO NOT MODIFY THIS FILE DIRECTLY
# This file is configured to deploy to:
#  - uat
#  - uat2

name: deploy
on:
  push:
    branches:
# ...
```

The api for this could be simple:

```ts
const preparedComment = `# This file is auto generated. Please see the ReadMe.md for instructions.
# DO NOT MODIFY THIS FILE DIRECTLY
# This file is configured to deploy to:
${ENVS_TO_DEPLOY_TO.map((e) => `  - ${e.name}\n`)}
`;

const pipeline = /*...*/;
pipeline.workflowFile.commentBefore = preparedComment;
```

This would need to be built-in, otherwise it'll invalidate/break the `diffProtection`, which is undesirable.

I believe the code for this wouldn't be too complicated. The only issue I see is the old version of the `yaml` that's being used has a different API for these functions. At this point, I believe switching to the newer `yaml` library would be trivial, but these changes in particular would use the parts of the API that have changed.

If interested, I could work up a PR.

kaizencc:
hi @giseburt! Adding in these comments would be helpful. I'm happy with upgrading yaml to a newer version as well. Ideally those would come in two separate PRs. If you're willing to contribute, that would be awesome!

giseburt:
@kaizencc Ok, the update to the new YAML was trivial, and can be found in #353.

I'll make a PR for the second part within a day or so.

PR #353, opened 2022-10-04
chore: Update `yaml` dependency to latest

giseburt:
Part 1 of 2 for #349

All test passed locally, no new tests appear to be needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
   "dependencies": {
     "decamelize": "^5.0.1",
     "fast-json-patch": "^3.1.1",
-    "yaml": "^1.10.2"
+    "yaml": "^2.1.2"
   },
   "bundledDependencies": [
     "decamelize",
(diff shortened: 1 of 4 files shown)


kaizencc (approved):
Thanks!

kaizencc:
@giseburt, you need to run `yarn build` locally and update the snapshots, I can't do it for you because your `main` branch is protected

giseburt:
Ok. I was about to but saw you were pushing code and don't want to conflict. I'll do that shortly.

giseburt:
@kaizencc Ok, should be good now.

I'm not sure why the test snapshots needed updated. This change shouldn't have triggered that, but I adjusted anyway.

BTW, unrelated issue: ` --updateSnapshot` shouldn't be on the `jest` command for testing ([here](https://github.com/cdklabs/cdk-pipelines-github/blob/main/.projen/tasks.json#L242)) - that will make any test with a snapshot _always_ pass  since it'll just update the snapshot and pass it.

kaizencc:
@giseburt it must be that yaml 2.0 handles undefined objects differently. Looks like previously, `container: undefined` would render as `container: null` and now it is omitted entirely. I think this is okay, given the experimental nature of this module, people should be fine that their `deploy.yml` files change in a insignificant way.

PR #354, opened 2022-10-04
feat: add yaml comments to deploy workflow

### LINKED ISSUE #349: Ability to set a top-of-file comment on deploy.yaml

giseburt:
Fixes #349


giseburt:
Had to force-push on my fork, sorry. Not sure why git pulled and didn't make a merge-commit before.

giseburt:
@kaizencc This should include the latest `main` as well as an update to `API.md` that I missed before.

README.md
```diff
@@ -53,24 +54,21 @@ called `MyStage` that includes CDK stacks for your app and you want to deploy it
 to two AWS environments (`BETA_ENV` and `PROD_ENV`):
 
 ```ts
-import { App } from 'aws-cdk-lib';
-import { ShellStep } from 'aws-cdk-lib/pipelines';
-import { GitHubWorkflow } from 'cdk-pipelines-github';
+import { App } from "aws-cdk-lib";
+import { ShellStep } from "aws-cdk-lib/pipelines";
+import { GitHubWorkflow } from "cdk-pipelines-github";
```
giseburt, line 59: My apologies, these changes were done by my VSCode running the eslint rules against the ts code blocks in the Markdown automatically when I saved. I can back these changes out if you prefer, just let me know.

src/yaml-file.ts
```diff
@@ -34,6 +34,16 @@ export class YamlFile {
    */
   private readonly patchOperations: JsonPatch[];
 
+  /**
+   * Internal storage for commentBefore accessors.
+   */
+  private _commentBefore?: string;
```
kaizencc, line 40:
```suggestion
  private commentBefore?: string;
```

Since these are private, I see no reason to underscore

src/yaml-file.ts
```diff
@@ -34,6 +34,16 @@ export class YamlFile {
...
+
+  /**
+   * Internal storage for commentAfter accessors.
+   */
+  private _commentAfter?: string;
```
kaizencc, line 45:
```suggestion
  private commentAfter?: string;
```

src/yaml-file.ts
```diff
@@ -87,13 +97,81 @@ export class YamlFile {
...
+   * ...
+   * ```
+   */
+  public set commentAtTop(v: string | undefined) {
+    this._commentBefore = v;
+  }
+
+  public get commentAtTop(): string | undefined {
+    return this._commentBefore;
+  }
```
kaizencc, line 128: i don't see a use-case for this getter. I think we can just get rid of this and the getter for `commentAtBottom`

test/yaml-file.test.ts
```diff
@@ -6,14 +6,18 @@ import { YamlFile } from '../src/yaml-file';
 import { withTemporaryDirectory } from './testutil';
 
 describe('patch', () => {
-  function patchTest(patches: JsonPatch[], initialObj: any, assertObj: any) {
+  function patchTest(patches: JsonPatch[], initialObj: unknown, assertObj: unknown, updateObj?: unknown) {
```
kaizencc, line 9: why unknown over any?

README.md
```diff
@@ -466,6 +446,36 @@ before the deployment can proceed:
...
+add boilerplate explaining the pipeline file as well as describe what's being
+deployed.
+
+```ts
+const pipeline = new GitHubWorkflow(/* ... */);
+
+pipeline.workflowFile.commentAtTop = `AUTOGENERATED FILE, DO NOT EDIT DIRECTLY!
+See ReadMe.md for instructions and explanation.`;
+
+pipeline.workflowFile.commentAtBottom = `THE END`;
```
kaizencc, line 461: I don't think its necessary to actually expose this to the user. I think we should just default to adding a comment at the top (and have no comment at the bottom). 

The only real use case for this is to say that this is an autogenerated file. I don't see any reason why users would want to modify this message in any way (or add anything to the bottom).

kaizencc (changes requested):
Hi @giseburt! My main comment is that I don't think this is a feature to be exposed to users. We should just default with a message and leave it at that. That should simplify this PR greatly as well.

src/yaml-file.ts
giseburt: The underscore is for disambiguating from the getter/setter.

The getter and setter were added out of habit for future expansion where there may be a desire to possibly alter the way that these values are stored, add hooks on when the values is set or read, etc.

These are unlikely in this case I suppose.

README.md
giseburt: I wasn't aware of any immediate use cases either, but figured it cost nothing to expose and doesn't pose a risk of invalidating the file. 🤷

test/yaml-file.test.ts
giseburt: Using `any` is much more dangerous than `unknown`. With `any` you can do anything with it, reverting to raw JavaScript behaviors (IOW no checks). With `unknown` it is saying that we don't want to disable checks, but we also aren't (yet) aware of the shape of this data.

If we do nothing with it directly but to pass it to other functions, then it makes little difference. But since code changes can later make a pass-through case into a use-it-now case one generally should avoid `any`.

Here's a more in-depth answer (not by me): https://dev.to/arikaturika/typescript-why-to-use-unknown-instead-of-any-41i8

giseburt: I should add that this change came from my configured [typescript-eslint](https://typescript-eslint.io/) static analysis rules. It's good practice, but I also generally avoid changing code unrelated to a PR in public PRs.

This is also what happened to the quotes in the markdown file. I have adjusted my local settings to be less aggressive when saving in non-local repos, sorry about that.

giseburt:
In my use case we have branch-specific stacks that are configured in the repo. I would like to be able to say both that this is autogenerated but also a human-readable (at-a-glance) message about what configuration this file represents, since it has to be committed when it changes. IOW, non-standard behavior I wouldn't expect this package to support directly.

I don't see how exposing this is harmful. Unlike the escape hatches, I don't believe this can cause a pipeline file to become malformed.

That said, a default value of a warning about the file being automatically generated would go a long way.

giseburt:
I rolled back the unnecessary changes to the ReadMe, the `any` → `unknown`, and removed the getter/setter, leaving just the direct-access `commentAtTop` or `commentAtBottom`.

I don't wish to disagree, but without the ability to actually set the comments this PR wouldn't fix anything for us, since we'd still be required to modify the file post-build and invalidate the `diffProtection` checks.

kaizencc:
Hi @giseburt! Answering a few of your comments here:

> I don't wish to disagree, but without the ability to actually set the comments this PR wouldn't fix anything for us, since we'd still be required to modify the file post-build and invalidate the diffProtection checks.

Disagreeing is fine; I'm more so trying to figure out the use cases for features we add. Not trying to be a dictator here.

>  In my use case we have branch-specific stacks that are configured in the repo. I would like to be able to say both that this is autogenerated but also a human-readable (at-a-glance) message about what configuration this file represents, since it has to be committed when it changes. IOW, non-standard behavior I wouldn't expect this package to support directly.

This is a valid use case :).

> I don't see how exposing this is harmful. Unlike the escape hatches, I don't believe this can cause a pipeline file to become malformed.

Exposing this isn't harmful, but if there isn't a use case for it it doesn't make sense to. It will just be additional noise. However, looks like we do have a use case as you've described.

----

Given this, here's my proposal:

- Let's only deal with a comment at the top. If there's a use case for bottom comments lets add that later.
- Let's have a basic default for the top comment.
- Let's have an exposed property that can change the top comment or remove it altogether. -- This should be your use case right?

How does that sound?

giseburt:
Those all sound good. As for the default comment, how does this sound:

```yaml
# AUTOMATICALLY GENERATED FILE, DO NOT EDIT MANUALLY.
# Generated by AWS CDK and [cdk-pipelines-github](https://github.com/cdklabs/cdk-pipelines-github)
```

We could remove the link, but I figured we should maybe try to indicate *what* generated it.

kaizencc:
@giseburt that sounds good. looking forward to getting this merged!

giseburt:
@kaizencc I think that last commit will do it. 👂

test/yaml-file.test.ts
```diff
@@ -32,15 +36,64 @@ describe('patch', () => {
...
   });
 
+  test('patch(p, v) can add to an existing array after update', () => {
+    patchTest(
+      [
+        JsonPatch.add('/first/second/array/-', '1'),
+        JsonPatch.add('/first/second/array/-', '2'),
+        JsonPatch.add('/first/second/array/1', '3'),
+      ],
+      { first: { second: { array: ['?'] } } }, // shound't use
```
kaizencc, line 46:
```suggestion
      { first: { second: { array: ['?'] } } }, // shouldn't use
```

kaizencc (approved):
Awesome @giseburt! I think this is a really great feature that we're adding to this library.

PR #365, opened 2022-10-10
fix: change `GitHubActionStep.jobStep` to `GitHubActionStep.jobSteps` (plural)

### LINKED ISSUE #363: Support multiple steps in `GitHubActionStep`
This basically means changing `GitHubActionStep.jobStep: JobStep` → `GitHubActionStep.jobSteps: JobStep[]`.

Would also like the ability to have an `if` clause on `GitHubActionStep` that overrides that of the `JobSettings`.

I have a PR teed up if this is acceptable.

giseburt:
Fixes #363 

Also adds `if` option to `GitHubActionStep` - committed separately so it can be backed out easily.

BREAKING CHANGE: `jobStep` has been changed to `jobSteps`.


test/stage-options.test.ts
```diff
@@ -216,31 +218,40 @@ test('can set pre/post github action job step', () => {
...
           uses: 'my-post-deploy-action@1.0.0',
           with: {
             'app-id': 4321,
             'secrets': 'secrets',
           },
-        },
+        }],
+        if: 'check on PostDeployAction only',
       })],
+      jobSettings: { if: 'check on Deploy MyStackXYZ only' },
```
kaizencc, line 245: can we make these `if`s into real if statements that would be accepted?

src/pipeline.ts
```diff
@@ -651,16 +651,14 @@ export class GitHubWorkflow extends PipelineBase {
       id: node.uniqueId,
       definition: {
         name: step.id,
-        ...this.jobSettings,
+        if: step.if === '' ? undefined : (step.if ?? this.jobSettings?.if),
```
kaizencc, line 654: I'm not sure if i like this; lets discuss here. Specifically, `jobs.if = ''` removing the `if` clause doesn't feel acceptable to me. First, we're assuming `this.jobSettings` will only have `if`. The property is meant to encompass all settings applied to all jobs in the workflow and might have more than `if` in the future.

Second, I question whether we need to support this use case here. `jobSettings` is meant to be applied to all jobs in the workflow, so `step.if === ''` is providing a magic value as the exception to that rule. I believe that if someone wants to apply an "exception", it should be done via escape hatches rather than provide a magic value that does the exception for them.

kaizencc (changes requested):
Hi @giseburt! Thanks for your contributions to this repository thus far! This PR looks good modulo a few comments. Thanks for the various adjacent fixes as well.

src/pipeline.ts
giseburt: 🤔 

Going to the [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds), since:
- All jobs in the pipeline directly or indirectly `needs` the `Build-*` step (the `*` is named by the `synth: ShellStep` [^1])
- AND In the absence of an `if` on a `job`, when any of the needed jobs fail then that `job` won't get run, effectively `if: 'success()'` ([`success()` docs](https://docs.github.com/en/actions/learn-github-actions/expressions#success))

Then the `this.jobSettings?.if` only needs to be applied to the `Build-*` synth step *anyway.*

Ok, so following that, and assuming we make that adjustment, then this code simply becomes `step.if` and `this.jobSettings?.if` is no longer referenced.

> First, we're assuming `this.jobSettings` will only have `if`.

This is true. Perhaps a private function called like `...this.renderJobSettings(),` that would then make the code:

```suggestion
        ...this.renderJobSettings(),
        if: step.if,
```

And in `jobForBuildStep` it would be:

```ts
    return {
      id: node.uniqueId,
      definition: {
        name: 'Synthesize',
        ...this.renderJobSettings(),
        if: this.jobSettings?.if,
        permissions: {
...
```

Or, we pass a parameter like `...this.renderJobSettings(true),` to render the job settings for the build step, including the `if`?

[^1]: Perhaps we should have a stable name for the build step so it can be referred to in cases like `${{ needs.Build-Synth.outputs.build_id }}`?

giseburt: @kaizencc I added commit to handle most of this, let me know what you think.

test/stage-options.test.ts
```diff
@@ -187,13 +187,14 @@ describe('job settings', () => {
...
-          if: 'github.repository == \'github/repo\'',
+          if: 'github.repository == \'github/repoB\'',
         },
       });
 
       app.synth();
 
-      expect(readFileSync(pipeline.workflowPath, 'utf-8')).toContain('if: github.repository == \'github/repo\'\n');
+      expect(readFileSync(pipeline.workflowPath, 'utf-8')).toContain('if: github.repository == \'another/repoA\'\n');
+      expect(readFileSync(pipeline.workflowPath, 'utf-8')).toContain('if: github.repository == \'github/repoB\'\n');
```
kaizencc, line 197: can we snapshot test this one so that we can look at the resulting pipeline changes and reason through if they are the same?

kaizencc (changes requested):
Hi @giseburt! I think the change from `jobStep` to `jobSteps` is complete and makes sense. My only comment there is that I'm sure there's a README update that needs to happen that is not in this PR. 

I suggest we split out the additional `if` work to a separate PR. That way we can stay true to what this PR says it is dealing with, and keep it contained to that. I'm not against modifying the behavior of `jobSettings`, so long as we confirm that there are no pipeline behavioral changes. Not sure if we are there yet. It might just be simpler to offer a way to override the existing if rather than removing it altogether.

test/stage-options.test.ts
giseburt: Sure.

Based on your other comment, it sounds like you'd prefer the `if` work to be in its own PR, yeah? How about this:
1. I'll back-out the `if` changes in this PR branch
2. Wait until this PR is merged
3. Then re-apply the `if` changes to a new branch off of the main branch and make a new issue+PR
4. Make this change there?

kaizencc: Yep exactly. Looks like `if` changes may be more complicated than this PR lets on, and I'd rather focus on it on a separate PR.

src/pipeline.ts
giseburt: Ok, so I tried what you have ☝️ , and it doesn't work: `Node with duplicate id: MyStack`

Ok, but to get to the point of what (I think) you're trying to do:
```ts
          installCommands: ['yarn'],
          commands: ['yarn build'],
        }),
      });

      const stageA = new Stage(app, 'MyStageA', {
      });

      pipeline.addStageWithGitHubOptions(stageA, {
        jobSettings: {
          if: 'success() && github.repository == \'github/repoA\'',
        },
      });

      const stageB = new Stage(app, 'MyStageB', {
      });

      pipeline.addStageWithGitHubOptions(stageB, {
        jobSettings: {
          if: 'success() && github.repository == \'github/repoB\'',
        },
      });

      app.synth();

      const workflowFileContents = readFileSync(pipeline.workflowPath, 'utf-8');
      expect(workflowFileContents).toMatchSnapshot();
```

In that case, run off of this branch with the latest code, the snapshot shows:
1. `Build-Build` and `Assets-FileAsset1` do NOT have `if` statements, so they'll always build
2. `MyStageA-MyStackA-Deploy` depends on `Build-Build` and `Assets-FileAsset1` and has `if: success() && github.repository == 'github/repoA'` so it'll work "as expected" [^1]
3. `MyStageB-MyStackB-Deploy` however depends on `Build-Build`, `Assets-FileAsset1`, AND `MyStageA-MyStackA-Deploy`, and has `if: success() && github.repository == 'github/repoB'`, so it won't ever build

This seems like something that should be configurable, but isn't at the moment. IIRC, the only way to prevent `MyStageB-MyStackB-Deploy` from depending on `MyStageA-MyStackA-Deploy` is with a wave, but there's no means to add GitHub options to a stage in a wave at the moment.

[^1]: If I understand it correctly, it will build even if the needed steps fail unless we add `success() &&` to the beginning of the `if`, so I did

giseburt: Ok, so I think the `if` PR should cover these cases, and since this is ballooning I'll remove the `if` code changes from this PR as discussed elsewhere and push.

giseburt:
@kaizencc Ok, I backed out the `if` changes and added some more to the ReadMe.

README.md
```diff
@@ -395,17 +399,28 @@ const pipeline = new GitHubWorkflow(app, 'Pipeline', {
...
+          'app-id': 1234,
+          'secrets': 'my-secrets',
+        },
       },
-    },
+      {
+        name: 'pre beta-deploy check',
+        run: npm run preDeployCheck
+      }
+    ]
```
kaizencc, line 423:
```suggestion
      {
        name: 'pre beta-deploy check',
        run: 'npm run preDeployCheck',
      },
    ],
```

kaizencc (approved):
Thanks for all your help thus far, @giseburt!

PR #368, opened 2022-10-11
fix: Fail asset publish job if cdk-asset fails

### LINKED ISSUE #367: Failing to publish assets, but job reports success
If publishing assets with cdk-assets fails, the job still suceeds. This is caused by running the command in a shell script:

https://github.com/cdklabs/cdk-pipelines-github/blob/main/src/pipeline.ts#L391

The shell script will still return successfully with exit code 0 even if the `cdk-assets` command fails.

The script should either `set -e` or `exec cdk-assets ...`.

christophgysin:
Fixes #367


kaizencc (approved):
Thanks @christophgysin!

PR #372, opened 2022-10-13
feat: Support runners with preconfigured credentials

### LINKED ISSUE #362: Support github runners which bring their own AWS credentials
Currently there are two ways how to pass credentials: 
* OIDC
* GitHub Secrets

In our current environment, we have runners backed by EC2 instances that already provide AWS credentials.

I would like to be able to choose neither of the above options, and let the workflow assume the runners already have credentials configured.

christophgysin:
This introduces a new property `awsCreds` to configure the AWS credential provider. The old properties `awsCredentials` and `gitHubActionRoleArn` are now deprecated.

The change is backwards compatible. To move to the new property, change:

    {
      ...
      awsCredentials: {
        awsAccessKeyId: 'MY_KEY_ACCESS_KEY_ID',
        secretAccessKey: 'MY_SECRET_ACCESS_KEY',
      },
    }

to:

    {
      ...
      awsCreds: AwsCredentials.fromGitHubSecrets({
        awsAccessKeyId: 'MY_KEY_ACCESS_KEY_ID',
        secretAccessKey: 'MY_SECRET_ACCESS_KEY',
      }),
    }

Or when using OpenID Connect, change:

    {
      ...
      gitHubActionRoleArn: 'myRoleArn',
    }

to:

    {
      ...
      awsCreds: AwsCredentials.fromOpenIdConnect({
        gitHubActionRoleArn: 'myRoleArn',
      }),
    }

Additionally, you can now configure the pipeline to not provide any AWS credntials, which can be useful when your runners already provide credentials themselves:

    {
      ...
      awsCreds: AwsCredentials.runnerHasPreconfiguredCreds(),
    }

Fixes #362


src/pipeline.ts
```diff
@@ -72,9 +72,9 @@ export interface GitHubWorkflowProps extends PipelineBaseProps {
    * Names of GitHub repository secrets that include AWS credentials for
    * deployment.
    *
-   * @default - `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`.
+   * @default - Get AWS credentials from GitHub secrets `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`.
    */
-  readonly awsCredentials?: AwsCredentialsSecrets;
+  readonly awsCredentials?: AwsCredentialsProvider;
```
christophgysin, line 77: This would be a breaking change. Please let me know if we can live with that, and I'll document it in the commit message.

christophgysin: Of course, I'll add it back and introduce a new property instead.

src/pipeline.ts
```diff
@@ -207,6 +211,26 @@ export class GitHubWorkflow extends PipelineBase {
     this.publishAssetsAuthRegion = props.publishAssetsAuthRegion ?? 'us-west-2';
   }
 
+  /**
+   * Parse AWS credential configuration from deprecated properties For backwards compatibility.
+   */
+  private getAwsCredentials(props: GitHubWorkflowProps) {
```
kaizencc, line 217: Now that providing aws creds is convoluted with deprecated apis and such, I think it's necessary to have some checks here to make sure that the user is not supplying more than one of `githubActionRoleArn`, `awsCredentials`, or `awsCreds`.

src/pipeline.ts
```diff
@@ -85,6 +94,8 @@ export interface GitHubWorkflowProps extends PipelineBaseProps {
    * you can utilize the `GitHubActionRole` construct to create a role for you.
    *
    * @default - GitHub repository secrets are used instead of OpenId Connect role.
+   *
+   * @deprecated Use `awsCreds` instead.
```
kaizencc, line 98:
```suggestion
   * @deprecated Use `awsCreds.fromOpenIdConnect()` instead.
```

src/pipeline.ts
```diff
@@ -68,11 +68,20 @@ export interface GitHubWorkflowProps extends PipelineBaseProps {
...
+   */
+  readonly awsCreds?: AwsCredentialsProvider;
+
   /**
    * Names of GitHub repository secrets that include AWS credentials for
    * deployment.
    *
    * @default - `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`.
+   *
+   * @deprecated Use `awsCreds` instead.
```
kaizencc, line 84:
```suggestion
   * @deprecated Use `awsCreds.fromGitHubSecrets()` instead.
```

kaizencc (changes requested):
Hi @christophgysin, this looks great! some minor nits here and there, and also please update the readme with these changes. Most of the readme examples should be updated to not use deprecated apis and we should list the new way of supplying preconfigured credentials as well.

kaizencc (approved):
Thanks @christophgysin! Really appreciate all your contributions thus far.

PR #377, opened 2022-10-15
feat: support waves and add better `if` control

### LINKED ISSUE #376: Better flow control: Ability to use Waves, set `if` on individual stages/steps
This continues from the discussion in #365 with the `if` control.

Looking at the docs for [`jobs.<job_id>.if`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif), it appears that:

1. The `if` statement, if present, can allow a step to build if dependent steps (in `jobs.<job_id>.needs`) fail, skip, or are cancelled.
2. The `if` statement, if omitted, is the same as `if: success()` [^1]
3. The `if` expression options that allow bypassing failure are: (Same reference.)
    - `always()` - ignores failure AND cancellation
    - `failure()` - ONLY on failure
    - `cancelled()` - ONLY if the workflow was cancelled
    - `success() || failure()` - unlike `always()` this allows cancellation of a step
4. If a previous step is skipped then the `if` skips dependent steps unless one of the above expressions is added, known issue: actions/runner#491
    - The best proposed solution for NOT skipping dependent steps, but also not running the if there's been a failure:
    ```yaml
        if: |
          always() &&
          !contains(needs.*.result, 'failure') &&
          !contains(needs.*.result, 'cancelled') &&
          <ACTUAL TEST INTENDED HERE>
    ```
5. Closely related is `jobs.<job_id>.continue-on-error` - which allows a job to act as if it succeeds even when it fails, making a job that cannot fail

1, 2, and 4 means that we only need the `if` statement on the build phase, and can skip `if` statements on publish steps. Depending on how 4 gets "fixed" either an `if` will need to be added with `!(failure() || cancelled())` (As [this workflow run](https://github.com/nrkno/gha-workflow-status-function-testing/actions/runs/784816362/workflow) demonstrates).

I see no value on adding (customized) `if` statements on publish jobs in the absence of a way to correlate the asset being deployed with some tangible asset name. Even then the value is dubious, since it'll likely invalidate the deployment since asset publish steps are shared by deployment steps that use those assets.

In a closely related issue, `needs` is used to handle dependencies, which the only way to make a stage NOT depend on another stage is with `pipeline.addWave()`, but that doesn't support GitHub options such as this `if` in the current code.

In the PR i have ready, this code works as expected (explained more below):

```ts
        installCommands: ['yarn'],
        commands: ['yarn build'],
      }),
    });

    const stageA = new GitHubStage(app, 'MyStageA', {
      jobSettings: {
        if: "success() && contains(github.event.issue.labels.*.name, 'deployToA')",
      },
    });

    const stageB = new GitHubStage(app, 'MyStageB', {
      jobSettings: {
        if: "success() && contains(github.event.issue.labels.*.name, 'deployToB')",
      },
    });

    // Make a wave to have the stages be parallel (not depend on each other)
    const wave = pipeline.addWave('MyWave');
    wave.addStage(stageA);
    wave.addStage(stageB);
```

This makes four jobs:
- `Build-Build` - with `if` set
- `Assets-FileAsset1` - needs `Build-Build` but doesn't have `if` set otherwise
- `MyWave-MyStageA-MyStackA-Deploy` - needs `Build-Build` and `Assets-FileAsset1` and has the correct `if` set
- `MyWave-MyStageB-MyStackB-Deploy` - also needs `Build-Build` and `Assets-FileAsset1` and has the correct `if` set
    - If it were `pipeline.addStage(stageA);` and `pipeline.addStage(stageB);` then there would be an additional needs value of `MyWave-MyStageA-MyStackA-Deploy`

Obviously there's a new construct `GitHubStage`, and a new `GitHubWorkflow.addWave()`. We can discuss technical details in the PR about how these work and if we want to maintain those names, etc.

[^1]: According to the docs [here](https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions):
    > You can use the following status check functions as expressions in if conditionals. A default status check of success() is applied unless you include one of these functions.

giseburt:
Fixes #376

Continuing the description there:

I've added `GitHubStage` which extends `Stage`, adding the properties of `AddGitHubStageOptions` to `StageProps`. `Stage` ignores them, and they quickly get lost in the conversion to a `StageDeployment`, so we also need to catch that conversion.

I made a subclass of `Wave` called `GitHubWave` since n both `wave.addStage()` and `pipeline.addStage()` (which is defined in [`pipeline_base`](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/pipelines/lib/main/pipeline-base.ts#L92-L97) as `this.addWave(stage.stageName).addStage(stage, options)`) the the `Stage` object itself is tossed.

`GitHubWave` takes the parent `GitHubWorkflow` as an additional parameter, and `GitHubWorkflow.addWave()` calls `new GitHubWave(id, this, options)`. `GitHubWave` only overrides `addStage()` to call `parent.addingStageFromWave()` with the unaltered `Stage` in addition to `super.addStage(stage, options)`.

I have also added `GitHubWave.addStageWithGitHubOptions()` to match `GitHubWorkflow.addStageWithGitHubOptions()` as well as `GitHubWorkflow.addGitHubWave()`, but those can likely go away.

The trick is the [JSII rules](https://aws.github.io/jsii/user-guides/lib-author/typescript-restrictions/) preventing "overriding or implementing a member using covariant parameter or return types" - which means we cannot make `addStage()` or `addWave()` take different parameters or have a different return type. However, we can accept and return classes that are descendants of those types, and use `instanceof` to catch them and get the additional information and functionality that they carry. So, `pipeline.addStage()` can accept a `GitHubStage` as a `Stage`, and `pipeline.addWave()` can return a `Wave` that's actually a `GitHubWave`. This allows us to maintain the same call structure but pass more information when we want.

How well that works in the other languages is TBD. If the JSII rules are sufficient, then it should work. In the least, JSII compiles it. 😄


kaizencc:
This seems like an opinionated PR -- I'll review it when I have time, but it won't be this week.

giseburt:
I appreciate it.

I'm very much open to discussion on this. 

I was partially aiming at enabling functionality and syntax found in the codebuild pipeline without having to use the escape hatches.

giseburt:
Hello @kaizencc,

I'd like to utilize these features (more importantly the Wave support), and would like to provide as much support as possible moving forward with this PR.

I'm open to discussion on how to move forward with this PR, and alter it as needed. It appears you may prefer to remove the `if` adjustments, perhaps moving them to another ticket?

Thank you!

giseburt:
Superseded by #490

PR #385, opened 2022-10-19
feat: Allow disabling diffProtection using string argument

### LINKED ISSUE #381: diffProtection can't be disabled with `--context` option
AFAIK it is not possible to pass typed context values using `--context`. See also https://github.com/aws/aws-cdk/issues/2204.

This makes it impossible to get past the check at https://github.com/cdklabs/cdk-pipelines-github/blob/main/src/pipeline.ts#L300-L301:
```ts
    const diffProtection = this.node.tryGetContext('cdk-pipelines-github:diffProtection') ?? true;
    if (diffProtection && process.env.GITHUB_WORKFLOW === this.workflowName) {
```
I suggest to also allow string values (`"false"`) to disable the check.

christophgysin:
Fixes #381


src/pipeline.ts
```diff
@@ -297,7 +297,7 @@ export class GitHubWorkflow extends PipelineBase {
 
     // GITHUB_WORKFLOW is set when GitHub Actions is running the workflow.
     // see: https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
-    const diffProtection = this.node.tryGetContext('cdk-pipelines-github:diffProtection') ?? true;
+    const diffProtection = JSON.parse(this.node.tryGetContext('cdk-pipelines-github:diffProtection') ?? true);
```
kaizencc, line 300: won't this accept any value set as context as `true`? I'm fine with `false` and `"false"` turning off diff protection, but that's it.

christophgysin: It behaves the same as before for values `true`, `false`, `0`, `1`, `null`, `undefined`, `'true'`. Behavior has changed for `'false'`.

For reference, I  ran it through a bunch of other types (empty string, non-empty string, array, object. Some of these would now return an error. None of these I would consider valid values, but feel free to specify the expected behavior, and I can implement it accordingly.

Of course I can implement full backwards compatibility, but I feel this would unnecessarily complicate the code. It seems highly unlikely anyone is relying on this behavior.

```ts
$ node
Welcome to Node.js v16.18.0.
Type ".help" for more information.
> [true, false, 0, 1, null, undefined, "true", "false", "", "foo", {}, []].map(value => {
...   const before = Boolean(value ?? true)
...   const after = () => {
...     try {
...       return Boolean(JSON.parse(value ?? true))
...     } catch (err) {
...       return `Error: ${err.message}`;
...     }
...   }
...   return {value, before, after: after()}
... })
[
  { value: true, before: true, after: true },
  { value: false, before: false, after: false },
  { value: 0, before: false, after: false },
  { value: 1, before: true, after: true },
  { value: null, before: true, after: true },
  { value: undefined, before: true, after: true },
  { value: 'true', before: true, after: true },
  { value: 'false', before: true, after: false },
  {
    value: '',
    before: false,
    after: 'Error: Unexpected end of JSON input'
  },
  {
    value: 'foo',
    before: true,
    after: 'Error: Unexpected token o in JSON at position 1'
  },
  {
    value: {},
    before: true,
    after: 'Error: Unexpected token o in JSON at position 1'
  },
  {
    value: [],
    before: true,
    after: 'Error: Unexpected end of JSON input'
  }
]

Comment (christophgysin) [reply src/pipeline.ts:300]: This would be the minimal change needed to get 'false' handled correctly and be fully backwards compatible:
```suggestion
    const contextValue = this.node.tryGetContext('cdk-pipelines-github:diffProtection');
    const diffProtection = contextValue === 'false' ? false : contextValue ?? true;
```

kaizencc: Thanks for the deep dive. I was slightly misunderstanding before, so it helps. Now, my comment is that I don't think we should throw an error for invalid strings like `foo`, we should return `true`. Moreover, the error message would be cryptic because we're using `JSON.parse` for a value that really isn't a json.

I like your code suggestion better. Will add in something similar and then approve.

mmutwiri85-dev (commented):
Okay I got that

PR #490, opened 2023-01-20
feat: waves for parallel builds with github options

### LINKED ISSUE #489: Ability to use Waves
*This is a paired-down version of #376*

Would like the ability to use Waves, as in the following modified test case:

```ts
        installCommands: ['yarn'],
        commands: ['yarn build'],
      }),
      jobSettings: {
        if: "contains(github.event.issue.labels.*.name, 'deployToA') || contains(github.event.issue.labels.*.name, 'deployToB')",
      },
    });

    // Make a wave to have the stages be parallel (not depend on each other)
    const wave = pipeline.addWave('MyWave', {
      pre: [
        new GitHubActionStep('PreWaveAction', {
          jobSteps: [
            {
              name: 'pre wave action',
              uses: 'my-pre-wave-action@1.0.0',
              with: {
                'app-id': 1234,
                'secrets': 'my-secrets',
              },
            },
          ],
        }),
      ],

      post: [
        new GitHubActionStep('PostWaveAction', {
          jobSteps: [
            {
              name: 'Checkout',
              uses: 'actions/checkout@v3',
            },
            {
              name: 'post wave action',
              uses: 'my-post-wave-action@1.0.0',
              with: {
                'app-id': 4321,
                'secrets': 'secrets',
              },
            },
          ],
        }),
      ],
    });

    const stageA = new GitHubStage(app, 'MyStageA', {
      jobSettings: {
        if: "success() && contains(github.event.issue.labels.*.name, 'deployToA')",
      },
    });

    const stageB = new GitHubStage(app, 'MyStageB', {
      gitHubEnvironment: 'test',
    });

    wave.addStage(stageA);
    wave.addStage(stageB);
```

Currently we lose the ability to use `GitHubActionStep` in `pre` and `post` as well as any `AddGitHubStageOptions` values not inherited from `StageProps`.

giseburt:
Fixes #489

Removed the `if`-specific code from #377 and rebased on top of the latest main.

With this PR it's possible to have Waves without losing GitHub-specific settings properties.


kaizencc (changes requested):
Hi @giseburt! Thanks for the PR, and sorry for the delay in getting back to you. I really appreciate your continued interest in this library! Can you do the following steps to help create an easier entrypoint into this PR:

- no stylistic changes -- it's cluttering the PR and makes it harder to find actual necessary changes. I'm not against it, but if you want to add them in they deserve to be in its own PR, with a way to enforce the same style in the future.
- a README update -- The PR title tells me that this will support waves, but nowhere in the tests do I see how that would work in practice. I would like to work backwards from the use case here, in that I'd like to see what a README entry for the new feature would look like, before I dive into the nitty gritty implementation.

giseburt:
Ok, @kaizencc, I updated the docs to include the new objects and how to use them.

src/pipeline.ts
```diff
@@ -256,6 +303,74 @@ export class GitHubWorkflow extends PipelineBase {
...
+  }
+
+  /**
+   * Support adding stages with GitHub options to waves - should ONLY be called internally.
+   *
+   * Use `pipeline.addWave()` and it'll call this when `wave.addStage()` is called.
+   *
+   * `pipeline.addStage()` will also call this, since it calls `pipeline.addWave().addStage()`.
+   */
+  public addingStageFromWave(
```
kaizencc, line 344: why is this `adding` rather than `add`?

src/pipeline.ts
```diff
@@ -777,6 +895,10 @@ export class GitHubWorkflow extends PipelineBase {
 
     return deps.map(x => x.uniqueId);
   }
+
+  private renderJobSettingParameters() {
+    return this.jobSettings;
+  }
```
kaizencc, line 901: okay, lets keep it, but probably not necessary

src/pipeline.ts
```diff
@@ -816,6 +938,68 @@ function snakeCaseKeys<T = unknown>(obj: T, sep = '-'): T {
...
 }
 
+/**
+ * Multiple stages that are deployed in parallel
+ *
+ * A `Wave`, but with addition GitHub options
+ *
+ * Create with `GitHubWorkflow.addWave()` or `GitHubWorkflow.addGitHubWave()`, do not construct directly
+ *
+ * @hideconstructor
```
kaizencc, line 948: why `@hideconstructor`? what does it do and why do we need it here. can't we make the constructor private if we don't want it public?

src/pipeline.ts
```diff
@@ -816,6 +938,68 @@ function snakeCaseKeys<T = unknown>(obj: T, sep = '-'): T {
...
+  // public addPre(...steps: Step[]) {
+  //   this.pre.push(...steps);
+  // }
+
+  // /**
+  //  * Add an additional step to run after all of the stages in this wave
+  //  */
+  // public addPost(...steps: Step[]) {
+  //   this.post.push(...steps);
+  // }
```
kaizencc, line 1000: oops?

kaizencc (changes requested):
Hi @giseburt! Some initial thoughts. I'm still mostly interested in seeing in visual form what a pipeline with waves looks like in GitHub Workflows.

src/pipeline.ts
giseburt: It's an internal call that happens in the middle of the external `addStage()` or `addWave()` - it's the common code for both. I suppose other names could be something like `completeAddingStageFromWave()`.

src/pipeline.ts
giseburt: It was a futile attempt at getting the JSII doc builder to hide the constructor. I also couldn't find a clean way to make it private. I'll remove this, it shouldn't have made it into the push.

src/pipeline.ts
giseburt: Well, I kept these here commented for a few reasons that probably don't hold water:

1. If we did decide to override them the signature is there
2. To indicate that they are there and still usable, but not currently overridden
3. I was sure that some test case would show that I needed to add some logic there. It never did happen.

In any case, I'll remove these. They're easy to restore.

README.md
```diff
@@ -500,6 +523,119 @@ pipeline.addStageWithGitHubOptions(new MyStage(this, 'Prod', {
...
+        with: {
+          'app-id': 4321,
+          'secrets': 'secrets',
+        },
+      },
+    ],
+  }),
+]);
+```
+
```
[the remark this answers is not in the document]
giseburt: I'll see what I can do. The one example I have has some internal names from the client so I will need to make some adjustments for public consumption.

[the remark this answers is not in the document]
giseburt: I added some screen shots. Hopefully they're clear enough, the setup depicted is rather large (and real).

src/pipeline.ts
giseburt: Renamed it to `add` anyway. :-)

src/pipeline.ts
giseburt: :+1:

giseburt:
Ok, @kaizencc, I believe I have made all the requested changes and answered all questions. I'll keep an eye out or any more you might have. Thank you!

src/pipeline.ts
```diff
@@ -816,6 +938,52 @@ function snakeCaseKeys<T = unknown>(obj: T, sep = '-'): T {
...
+    super(id, props);
+  }
+
+  /**
+   * Add a Stage to this wave
+   *
+   * It will be deployed in parallel with all other stages in this
+   * wave.
+   */
+  public addStage(stage: Stage, options: AddStageOpts = {}) {
```
kaizencc, line 965: Is there anyway where we can have `addStage` call `addStageWithGitHubOptions` or the other way around. I don't like having two very similar methods co-existing. If we ever want to change one of the methods, we'd have to remember to change the other

test/stage-options.test.ts
```diff
@@ -202,56 +217,281 @@ test('can set pre/post github action job step', () => {
...
+        },
+        'MyWave-MyStageB-MyStackB-Deploy': {
+          if: stageBOptions.jobSettings?.if,
+        },
+      },
+    });
+  });
+});
+
+test('stages in pipeline works with `if`', () => {
```
kaizencc, line 444: is this test still relevant in the PR? How is it succeeding?

test/stage-options.test.ts
```diff
@@ -202,56 +217,281 @@ test('can set pre/post github action job step', () => {
...
+        },
+        'MyWave-MyStageB-MyStackB-Deploy': {
+          if: stageBOptions.jobSettings?.if,
+        },
+      },
+    });
+  });
+});
+
+test('github stages in waves works', () => {
```
kaizencc, line 346: can we add a test for throwing an error if you try to add a stage after it is built?

kaizencc (changes requested):

src/pipeline.ts
```diff
@@ -256,6 +303,74 @@ export class GitHubWorkflow extends PipelineBase {
...
+  }
+
+  /**
+   * Support adding stages with GitHub options to waves - should ONLY be called internally.
+   *
+   * Use `pipeline.addWave()` and it'll call this when `wave.addStage()` is called.
+   *
+   * `pipeline.addStage()` will also call this, since it calls `pipeline.addWave().addStage()`.
+   */
+  public addStageFromWave(
```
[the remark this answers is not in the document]
giseburt: It's `public` since it's called from inside the class `GitHubWave`. I'll try `@internal`.

src/pipeline.ts
giseburt: Makes sense. I'll have `addStage` call `this.addStageWithGitHubOptions(stage, options)` - no code duplication there.

src/pipeline.ts
```diff
@@ -816,6 +938,52 @@ function snakeCaseKeys<T = unknown>(obj: T, sep = '-'): T {
...
 }
 
+/**
+ * Multiple stages that are deployed in parallel
+ *
+ * A `Wave`, but with addition GitHub options
+ *
+ * Create with `GitHubWorkflow.addWave()` or `GitHubWorkflow.addGitHubWave()`, do not construct directly
+ */
+export class GitHubWave extends Wave {
```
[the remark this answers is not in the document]
giseburt: We do want them to *use* them directly, we just don't want them to construct them directly, and from there you need to be able to place them in a typed variable. From the code I'm using to test it, for example, I conditionally use a Wave:

```ts
const pipelineOrWave: GitHubWorkflow | GitHubWave | undefined;
```

src/pipeline.ts
```diff
@@ -175,8 +214,16 @@ export class GitHubWorkflow extends PipelineBase {
...
+  private readonly stackProperties: Record<
+  string,
+  {
+    environment: AddGitHubStageOptions['gitHubEnvironment'];
+    capabilities: AddGitHubStageOptions['stackCapabilities'];
+    settings: AddGitHubStageOptions['jobSettings'];
+  }
+  > = {};
   private readonly jobSettings?: JobSettings;
+  private builtGH = false; // cannot be `built` since that's defined in the parent as private
```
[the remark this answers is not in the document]
giseburt: The parent does this check [here](https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/pipelines/lib/main/pipeline-base.ts#L115-L117) in `PipelineBase.addWave()`. We can't use `this.built` since it's declared as `private` in `PipelineBase`, and we need to block the user from adding a Wave after the pipeline has built.

We can't use the provided `PipelineBase.addWave()` since it'll make a new `Wave`, so we override it.

I'm adding the comment:
```ts
  // in order to keep track of if this pipeline has been built so we can
  // catch later calls to addWave() or addStage()
```

src/stage-options.ts
```diff
@@ -53,4 +15,9 @@ export enum StackCapabilities {
...
   /** Acknowledge your stack contains one or more macros */
   AUTO_EXPAND = 'CAPABILITY_AUTO_EXPAND',
-}
\ No newline at end of file
+}
+
+/**
+ * Options to pass to `addStageWithGitHubOpts`.
+ */
+export interface AddGitHubStageOptions extends AddStageOpts, GitHubCommonProps {}
```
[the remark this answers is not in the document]
giseburt: Move `AddGitHubStageOptions` to `github-common.ts`? I'm going to assume that's what you meant, but in case you didn't...

src/pipeline.ts
```diff
@@ -816,6 +938,52 @@ function snakeCaseKeys<T = unknown>(obj: T, sep = '-'): T {
   return result as any;
 }
 
+/**
```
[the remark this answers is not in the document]
giseburt: Moved to `wave.ts`

test/stage-options.test.ts
giseburt: Yes. I modified it to track the current behavior. It appears that `if` added to different stages tracks those stages properly.

test/stage-options.test.ts
giseburt: :+1:

giseburt:
Ok, @kaizencc - I *think* I got all the changes you asked for.

vincentgna:
@kaizencc any update on this?

src/pipeline.ts
```diff
@@ -256,6 +268,76 @@ export class GitHubWorkflow extends PipelineBase {
...
+   * declare const pipeline: GitHubWorkflow; // assign pipeline a value
+   *
+   * const wave = pipeline.addWave('MyWave');
+   * wave.addStage(new MyApplicationStage(this, 'Stage1'));
+   * wave.addStage(new MyApplicationStage(this, 'Stage2'));
+   * ```
+   */
+  public addWave(id: string, options?: WaveOptions): Wave {
+    return this.addGitHubWave(id, options);
+  }
```
kaizencc, line 288: Hmm we probably just want to direct users to use `addGitHubWave` instead of `addWave` entirely, right? It feels confusing to have both `addWave` and `addGitHubWave` as part of the API.

I think we should consider either throwing an error when you use `addWave` and explaining that you should be using `addGitHubWave`, or at the very least, providing a warning (via `Annotations` construct) that you should be using `addGitHubWave`.

src/wave.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Multiple stages that are deployed in parallel
+ *
+ * A `Wave`, but with addition GitHub options
+ *
+ * Create with `GitHubWorkflow.addWave()` or `GitHubWorkflow.addGitHubWave()`, do not construct directly
+ */
+
+export class GitHubWave extends Wave {
+  constructor(
```
kaizencc, line 15:
```suggestion
  public constructor(
```

src/wave.ts
```diff
@@ -0,0 +1,49 @@
...
+import { AddStageOpts, StageDeployment, Wave, WaveProps } from 'aws-cdk-lib/pipelines';
+import { AddGitHubStageOptions } from './github-common';
+import { GitHubWorkflow } from './pipeline';
+
+/**
+ * Multiple stages that are deployed in parallel
+ *
+ * A `Wave`, but with addition GitHub options
+ *
+ * Create with `GitHubWorkflow.addWave()` or `GitHubWorkflow.addGitHubWave()`, do not construct directly
```
kaizencc, line 11:
```suggestion
 * Create with `GitHubWorkflow.addWave()` or `GitHubWorkflow.addGitHubWave()`.
 * You should not have to instantiate a GitHubWave yourself.
```

This should be documented on the constructor also.

kaizencc (changes requested):
Hi again @giseburt. Sorry for the delayed response. This is pretty much good to go and I'm prepared to approve tomorrow. I just have a few typo fixes and one comment about having both `addWave` and `addGitHubWave` as viable APIs, it might be confusing which one to use. Once that's addressed we're good!

src/pipeline.ts
giseburt: Unfortunately I spent a while trying to do this one to determine that it cannot be done, at least not without wrecking other things.

`PipelineBase.addStage` calls `addWave` internally. We'd have to rewrite `addStage` as well. I don't think it's worth it.

https://github.com/aws/aws-cdk/blob/main/packages/%40aws-cdk/pipelines/lib/main/pipeline-base.ts#L92-L97

I feel like the real solution would be upstream (adjust `PipelineBase`, etc, but I can't determine the best route at the moment.)

giseburt:
Ok, I once more believe I have all the changes you requested in, or have a good reason not to (and commented to that elsewhere).

kaizencc (approved):
Yay!

PR #518, opened 2023-02-20
feat: specify name and/or url in gitHubEnvironment

tj-harris:
Extended typing for `AddGitHubStageOptions.githubEnvironment` to allow for specifying an environment url. 

No changes to the underlying code were required.

Added a snapshot test to verify that the YAML file is created in accordance with  https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-environment-name-and-url

BREAKING CHANGE: `gitHubEnvironment` is now typed as a `GitHubEnvironment` and not `string`.


src/pipeline.ts
```diff
@@ -508,7 +529,7 @@ export class GitHubWorkflow extends PipelineBase {
       params['role-arn'] = resolve(stack.executionRoleArn);
     }
     const assumeRoleArn = stack.assumeRoleArn ? resolve(stack.assumeRoleArn) : undefined;
-
+
```
kaizencc, line 532:
```suggestion

```

src/pipeline.ts
```diff
@@ -520,8 +541,8 @@ export class GitHubWorkflow extends PipelineBase {
           idToken: this.awsCredentials.jobPermission(),
         },
         ...(this.stackProperties[stack.stackArtifactId]?.environment ? {
-          environment: this.stackProperties[stack.stackArtifactId].environment,
-        } : {}),
+            environment: this.stackProperties[stack.stackArtifactId].environment,
+          } : {}),
```
kaizencc, line 545:
```suggestion
          environment: this.stackProperties[stack.stackArtifactId].environment,
        } : {}),
```

src/stage-options.ts
```diff
@@ -46,11 +46,11 @@ export interface AddGitHubStageOptions extends AddStageOpts {
  */
 export enum StackCapabilities {
   /** Acknowledge your stack includes IAM resources */
-  IAM = 'CAPABILITY_IAM',
+  IAM = "CAPABILITY_IAM",
```
kaizencc, line 49: can we go back to single quotes pls

test/stage-options.test.ts
```diff
@@ -43,6 +43,33 @@ describe('github environment', () => {
...
+
+      new Stack(stage, 'MyStack');
+
+      pipeline.addStageWithGitHubOptions(stage, {
+        gitHubEnvironment: { name: 'test', url: 'test.com'},
+      });
+
+      app.synth();
+
+      expect(readFileSync(pipeline.workflowPath, 'utf-8')).toMatchSnapshot();
```
kaizencc, line 69: Lets not do a full snapshot here? Can we just assert that the correct environment is in the ensuing file?

src/stage-options.ts
```diff
@@ -18,7 +18,7 @@ export interface AddGitHubStageOptions extends AddStageOpts {
    *
    * @default - no GitHub environment
    */
-  readonly gitHubEnvironment?: string;
+  readonly gitHubEnvironment?: string | GithubEnvironment;
```
kaizencc, line 21: I don't think we are allowed to use union types here. I wonder how this got past jsii, and I suspect it probably didn't. At any rate, I think this is one of those cases where we thank our lucky stars we're an "experimental" module still, and break the API. What we want is something like this:

```ts
export interface GitHubEnvironment: {
  readonly name: string;
  readonly url?: string;
}
```

And then users use it like:

```ts
githubEnvironment: { name: 'blah' };
```

or 

```ts
gitHubEnvironment: { name: 'blah', url: 'blah.com' };
```

Then, just for completeness sake, we probably want to have a quick rendering step to make sure that `GitHubEnvironments` with only a name are formatted the way GitHub wants:

```ts
// here we can have a union type. Only in public APIs we cannot, due to jsii limitations.
renderGitHubEnvironment(env: GitHubEnvironment): string | GitHubEnvironment {
  if (env.url === undefined) {
    return: env.name;
  }
  return env;
}

**Review by kaizencc — CHANGES_REQUESTED** [review]
Hi @tj-harris, thanks for the contribution! The build is failing because I think you missed running (or committing the output of) `yarn build`.

File: src/stage-options.ts (hunk 5 above)
Comment (tj-harris) [reply src/stage-options.ts:21]: Awesome, I always forget about the JSII limits on typescript.

File: test/stage-options.test.ts [hunk 6]
```diff
@@ -53,6 +53,60 @@ describe('github environment', () => {
...
+      });
+
+      app.synth();
+
+      expect(readFileSync(pipeline.workflowPath, 'utf-8')).toMatch(/.*environment:\s+name: test\s+url: test\.com.*/m);
+    });
+  });
+
+
+  test('can specify one github environment with url', () => {
```
kaizencc, line 84: this test looks like a duplicate of the one above @tj-harris

kaizencc (changes requested):
looks good @tj-harris. Can you confirm that you didn't mean to add the duplicated test? then this should be ready to go :)

test/stage-options.test.ts
tj-harris: They are technically different but I see now that having two is redundant as the multiple environment test below covers the name only specification of a github environment. I will consolidate the tests.

tj-harris: Yeah straight up duplicated, removing the extra test.

kaizencc (approved):
Thanks @tj-harris!

PR #603, opened 2023-05-16
Add CFN parameter passing for Github Secrets and Environments

rdesmond1991:
**Problem**: A user wants to use Github Environment's Secrets or Environment variables in their CFN Stack
**Solution**: Add Stack Parameters to GithubStageParams and add them to the deployment step.

* Adds StackParameter and StackParameterType interface and enum respectively
* Adds formatStackParameter to translate type to github string

Based off of: https://github.com/aws-actions/aws-cloudformation-github-deploy

Fixes #


rdesmond1991:
Hi all, wanted to draw your attention to a few decisions made here.

1. Decided to abstract away env and secrets since they are explicitly part of the Github Environment feature-set
2. Decided not to abstract other possibilities such as github events, as seen [here](https://github.com/aws-actions/aws-cloudformation-github-deploy). These events are more freeform and abstracting all of it away means we need to keep up with github adding features. Therefore PLAIN_TEXT would be used to enter the syntax shown in the `aws-cloudformation-github-deploy` to access anything extra.

An alternative is to leave all in plain text and have the consumer write the Github syntax. Since the purpose of a GithubStage is to make use of the Github Environment's specifics, such as Secrets, Variables, and Approval Steps, I opted to have Secrets and Variables be abstracted out. A drawback here is that if Github decided to change their syntax, this package would need to change, though that would be true for anything in this package.

Please let me know if it should be changed.

rdesmond1991:
@kaizencc would you be able to assist here? Thank you!

src/github-common.ts
```diff
@@ -2,6 +2,28 @@ import { AddStageOpts } from 'aws-cdk-lib/pipelines';
...
+  SECRET,
+  ENV_VARIABLE,
+  PLAIN_TEXT,
+}
+
+export interface StackParameter {
+  /**
+   * StackParameterType
+   *
+   * required
```
kaizencc, line 15:
```suggestion
```

src/github-common.ts
```diff
@@ -2,6 +2,28 @@ import { AddStageOpts } from 'aws-cdk-lib/pipelines';
...
+   */
+  readonly type: StackParameterType;
+
+  /**
+   * Parameter value
+   *
+   * required
```
kaizencc, line 22:
```suggestion
```

kaizencc (changes requested):
Thanks for the contribution @rdesmond1991 and sorry for the long wait for a review. This looks good to me save for a few formatting nits

kaizencc:
Hi @rdesmond1991, I am closing this PR for a few reasons:

1) No action on this for months
2) The PR was cut from the `main` branch from an `org`. This means that we cannot have automatic merges for this PR, which severely hampers our automation. Ideally a PR comes from your fork, and not your `main` branch.

However, I recall being fairly happy with the state of this PR way back when. I would be happy to see a new PR fixing the above issues as well as addressing the comments from my previous review. 

Cheers

PR #846, opened 2024-01-18
feat: configure additional job settings for docker asset publishing jobs

corymhall:
Currently if you want to change any of the job settings for the docker asset publishing jobs you can use escape hatches. This PR adds a new property `dockerAssetJobSettings` which allow you to provide additional settings when you create the workflow so you don't have to use escape hatches.

Currently I've only implemented support for two things:
1. `setupSteps` - These are additional steps to run just prior to building/publishing the image. The use case I've run into for this is the need to setup docker `buildx` prior to running the build.
2. `permissions` - Allow setting/overriding permissions for docker asset steps. For example, when interacting with GitHub packages you have to set the `packages: read` permission.

Fixes #


kaizencc (changes requested):
Thanks @corymhall -- can you add an example somewhere in the readme?

corymhall:
> Thanks @corymhall -- can you add an example somewhere in the readme?

@kaizencc done!

kaizencc (commented):
Thanks @corymhall :)

kaizencc (approved):

kaizencc (approved):

kaizencc:
@corymhall build failing? i can look at it later if you'd like

PR #854, opened 2024-01-25
feat: authenticate with ghcr docker credentials

### LINKED ISSUE #843: Support ghcr docker registry
Currently the `DockerCredential` class supports a custom registry option, but the [implementation](https://github.com/cdklabs/cdk-pipelines-github/blob/07c2661aca38bb2e67685e7f4f0824fcc7b060b2/src/pipeline.ts#L796-L799) expects everything to be stored in a secret.

To configure the ghcr.io (GitHub packages) registry you do this

```yml
      - uses: docker/login-action@v2
        with:
          registry: ghcr.io
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}
```

I think there should be a new method (`customRegistryV2`?) that puts the responsibility on the user to provide the entire string rather than just the secret key.

```ts
DockerCredential.customRegistryV2('ghcr.io', {
  username: '${{ github.actor }}',
  password: '${{ secrets.GITHUB_TOKEN }}',
})

```

corymhall:
This adds the ability to authenticate with the GitHub Container Registry (GHCR) in docker asset publishing steps.

closes #843

### DIFF
--- a/API.md
+++ b/API.md
@@ -340,6 +340,9 @@ const pipeline = new GitHubWorkflow(app, 'Pipeline', {
     // Authenticate to ECR
     DockerCredential.ecr('<account-id>.dkr.ecr.<aws-region>.amazonaws.com'),
 
+    // Authenticate to GHCR
+    DockerCredential.ghcr(),
+
     // Authenticate to DockerHub
     DockerCredential.dockerHub({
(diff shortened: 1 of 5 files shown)


kaizencc (approved):
Nice! No breaking changes either it seems.

Issue #820, opened 2023-12-19
Add support for aws-cn partition


mbergkvist:
I'm working on support for deploying to aws-cn, the wip-pr is here: https://github.com/mbsoftworks/cdk-pipelines-github/pull/1

It works for my use case (though there might be more required changes for other use cases) but I need some help to remove the hard-coded `aws-cn` and `.amazonaws.com.cn` changes (https://github.com/mbsoftworks/cdk-pipelines-github/pull/1/commits/d2ed0187a543b75d5a70408e7a71139eb44c10e9). Any suggestion on what needs to be done to fix those?

kaizencc:
@mbergkvist -- the ask is to remove the hardcoded `aws` partition to instead be possibly `aws-cn` based on some information right? I am open to it... we can either have it be a well-known environment variable if you want to modify the partition, or we can somehow get the info from the stack environment. I'm happy to start with env variable because that sounds easier, at the expense of some user experience, but at least it's possible :).

This module isn't really maintained for new feature requests. I am happy to review any community contributions on this, but do not have time to do so on my own.

mbergkvist:
@kaizencc 

> I am happy to review any community contributions on this, but do not have time to do so on my own.

I'm happy to work on this, but I'd need some guidance. I'm not too familiar with the inner workings of the CDK framework.

> we can either have it be a well-known environment variable if you want to modify the partition, or we can somehow get the info from the stack environment

An environment variable could work, but as you say, it comes with a cost for the user experience. The alternative is to get the infomration from the `StackDeployment` parameter, but unfortunately it doesn't carry the partition (in the long-term, should there be a PR in aws-cdk-lib pipelines?), but perhaps we can draw some conclusion based on the region? E.g. if the region begins with `cn-` then we set the partition to `aws-cn`.

The second part in my PR, the replacement of `.amazonaws.com` with `.amazonaws.com.cn`, is an hack I'm not particular proud of and I think it should be changed upstream so correct URL's are created from the start for the assets. Nevertheless, we can start with adding the change behind the same region logic as above?

I think the changes in `src/oidc-provider.ts` is legit, should I create a separate PR for those?

mbergkvist:
@kaizencc I created a separate PR (#868) for replacing the hard coded aws partition in `GitHubActionRole` with AWS:Partition. I think that can be merged without the other changes necessary for aws-cn support.

mbergkvist:
[...]

PR #868, opened 2024-02-06
fix: make existingGitHubActionsProvider aware of AWS partitions

mbergkvist:
Replace hard coded 'aws' partition with a reference to the `AWS::Partition` in the current stack. This makes `GitHubActionRole` work for e.g. `aws-cn` too.

There are more work that needs to be done to get full aws-cn support (see #820), but this is a general fix and so could be merged independently.

### DIFF
--- a/API.md
+++ b/API.md
@@ -743,7 +743,7 @@ You can do this manually in the console, or create a separate stack that uses th
 You must `cdk deploy` once (with your normal AWS credentials) to have this role created for you.
 
 You can then make note of the role arn in the stack output and send it into the Github Workflow app via
-the `gitHubActionRoleArn` property. The role arn will be `arn:aws:iam::<accountId>:role/GithubActionRole`.
+the `gitHubActionRoleArn` property. The role arn will be `arn:<partition>:iam::<accountId>:role/GithubActionRole`.
 
 > [https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services)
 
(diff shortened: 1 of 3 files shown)


kaizencc:
Looks like the build is failing + some additional context.

I _think_ this is okay because the OIDC provider construct is meant to be deployed with traditional `cdk deploy`. However, the CDK stacks that are deployed via the pipeline do not get the traditional treatment. Instead, we are currently resolving a few magic strings:

https://github.com/cdklabs/cdk-pipelines-github/blob/6a913cd66b37f1f502e0ce78d016680b168782be/src/pipeline.ts#L604-L610

One of which is a hardcoded partition. Again, I think the changes here is fine, but it might be worth a full deploy just to check.

mbergkvist:
I got the following error messages from the failing build

> Files were changed during build (see build log). If this was triggered from a fork, you will need to update your branch.

I've rebased the branch to the main branch, can we see if this makes the GH action pass?

kaizencc:
Hi @mbergkvist I believe the remedy here is for you to run `yarn build` locally and push the result

mbergkvist:
@kaizencc I get the following error now when building locally

> [2024-02-09T06:49:00.369] [ERROR] jsii/package-info - Unable to find a JSII dependency named "aws-cdk-lib" as "^2.80.0". If you meant to include a non-JSII dependency, try adding it to bundledDependencies instead.

I'm following the steps in the pipeline

```shell
yarn install --check-files
npx projen build
```

Also trying `yarn build`, but I get the same error regardless. Not sure what I'm missing here? 🤷

kaizencc:
A couple of things @mbergkvist 

- this means that your aws-cdk-lib file is older than 2.80.0. you need to run `yarn install` or `npx projen` to update.
- i would do this for you but you've cut a PR from a org-level repository `mbsoftworks` so I do not have permissions. next time, I'd appreciate it if you contribute via your personal fork.
- what you're missing is a one line `API.md` update. If you'd like, I can open the same PR with the correct file diff, but then you wouldn't get direct credit for the PR. up to you though.

mbergkvist:
@kaizencc I did `yarn install` and `npx projen build` but it was still failing. But something must've been broken locally, I did a clean checkout of the repo and now the build works as expected 🤷‍♂️ 

The PR is updated with the changes in `API.md`.

Thanks for your support and patience!

PR #903, opened 2024-03-04
feat: expose concurrency at GitHub workflow level

zezke:
This PR adds a concurrency setting on workflow level, preventing multiple workflows run in parallel which our pipeline does not like.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1422,9 +1422,9 @@ camelcase@^6.2.0, camelcase@^6.3.0:
   integrity sha512-Gmy6FhYlCY7uOElZUSbxo2UCDH8owEk996gkbrpsgGtrJLM3J7jGxl9Ic7Qwwj4ivOE5AWZWRMecDdF7hqGjFA==
 
 caniuse-lite@^1.0.30001587:
-  version "1.0.30001591"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001591.tgz#16745e50263edc9f395895a7cd468b9f3767cf33"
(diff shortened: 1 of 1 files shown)


test/github.test.ts
```diff
@@ -360,6 +360,30 @@ test('example app', () => {
...
+        group: '${{ github.workflow }}-${{ github.ref }}',
+      },
+    });
+
+    app.synth();
+
+    console.log(github.workflowPath);
+    console.log(github.workflowPath);
+    console.log(github.workflowPath);
+    console.log(github.workflowPath);
```
kaizencc, line 381: Is this part of the concurrency test? how? please document otherwise next time i look at this i will laugh and then delete...

src/workflows-model.ts
```diff
@@ -956,3 +956,17 @@ export interface PublicOptions { }
...
+ */
+export interface ConcurrencyOptions {
+  /**
+   * The concurrency group to use for the job.
+   */
+  readonly group: string;
+  /**
+   * Conditionally cancel currently running jobs or workflows in the same concurrency group
+   */
+  readonly cancelInProgress?: boolean;
```
kaizencc, line 971: add a default here. probably `false`?

src/workflows-model.ts
```diff
@@ -956,3 +956,17 @@ export interface PublicOptions { }
  */
 export interface StatusOptions { }
 //#endregion
+
+/**
+ * Concurrency options
```
kaizencc, line 961: Can we link to the relevant github documentation please

kaizencc (changes requested):
Thanks for the PR @zezke! I have a few comments, also I would love for you to add this feature to the readme.

PR #905, opened 2024-03-05
feat: concurrency at GitHub workflow level

### LINKED ISSUE #915: Can't set concurrency
Currently there isn't much customisation available for job settings, limited to just "if" out of the optional settings. I want to be able to enable concurrency for a project, and figure it makes sense for the other job settings to be available too. 

I've made a PR with a possible implementation: https://github.com/cdklabs/cdk-pipelines-github/pull/914

zezke:
This PR adds a concurrency setting on workflow level, preventing multiple workflows run in parallel which our pipeline does not like. It also addressed the comments by @kaizencc in [the incorrect older PR](https://github.com/cdklabs/cdk-pipelines-github/pull/903).

Fixes #915

### DIFF
--- a/API.md
+++ b/API.md
@@ -44,6 +44,7 @@ Workflows.
       - [Manual Approval Step](#manual-approval-step)
     - [Pipeline YAML Comments](#pipeline-yaml-comments)
     - [Common Configuration for Docker Asset Publishing Steps](#common-configuration-for-docker-asset-publishing-steps)
+    - [Workflow Concurrency](#workflow-concurrency)
   - [AWS China partition support](#aws-china-partition-support)
   - [Tutorial](#tutorial)
   - [Not supported yet](#not-supported-yet)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -645,6 +646,32 @@ const pipeline = new GitHubWorkflow(app, 'Pipeline', {
 app.synth();
 ```
 
+### Workflow Concurrency
+
+If you want to prevent your workflow from running in parallel you can specify the concurrenct at workflow level. Below is an example of a workflow that will not run in parallel and where a running workflow will be cancelled in favor of the more recent one. The [GitHub docs](https://docs.github.com/en/actions/using-jobs/using-concurrency) provide further details on this.
```
kaizencc, line 651:
```suggestion
If you want to prevent your workflow from running in parallel you can specify the concurrency at workflow level. 
Below is an example of a workflow that will not run in parallel and where a running workflow will be cancelled in favor of the more recent one.
The [GitHub docs](https://docs.github.com/en/actions/using-jobs/using-concurrency) provide further details on this.
```

src/workflows-model.ts
```diff
@@ -956,3 +956,19 @@ export interface PublicOptions { }
...
+/**
+ * Concurrency options at workflow level
+ *
+ * @see https://docs.github.com/en/actions/using-jobs/using-concurrency
+ */
+export interface ConcurrencyOptions {
+  /**
+   * The concurrency group to use for the job.
+   */
+  readonly group: string;
```
kaizencc, line 969:
```suggestion
  readonly group: string;

```

src/workflows-model.ts
```diff
@@ -956,3 +956,19 @@ export interface PublicOptions { }
...
+  /**
+   * Conditionally cancel currently running jobs or workflows in the same concurrency group. Default: false
```
kaizencc, line 971:
```suggestion
   * Conditionally cancel currently running jobs or workflows in the same concurrency group.
   *
   * @default false
```

kaizencc (commented):
Thanks @zezke. Following the conversation on #914 I want to accept this change.

kaizencc (approved):

PR #906, opened 2024-03-05
feat: gha step can request access to id-token

zezke:
I would like to be able to execute AWS CLI commands in the post phase of a stage. In my case I obtain some CloudFormation outputs and provide them to the frontend.

Unfortunately it is not possible right now since GitHubActionSteps never have access to the `id-token`. Please provide feedback if you do not agree with the naming.


README.md
```diff
@@ -441,7 +441,9 @@ If you want to call a GitHub Action in a step, you can utilize the `GitHubAction
 
 The `jobSteps` array is placed into the pipeline job at the relevant `jobs.<job_id>.steps` as [documented here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idsteps).
 
-In this example,
+You can use the `useGitHubActionRole` property if you need read access to the id-token, like you need when interacting with AWS.
```
kaizencc, line 444: Can you enter into the README a concrete example of when a user would need/want to use `useGitHubActionRole`?

src/steps/github-action-step.ts
```diff
@@ -11,6 +11,13 @@ export interface GitHubActionStepProps {
    * Environment variables to set.
    */
   readonly env?: Record<string, string>;
+
+  /**
+   * Indicates if this step uses the GitHub action role and thus should have access to it. When set
+   * to true, the step will get the 'id-token: read' permission.
+   * @default false
```
kaizencc, line 18:
```suggestion
   * to true, the step will get the 'id-token: read' permission.
 
   * @default false
```

kaizencc (commented):
Just one comment here @zezke

kaizencc:
Closing for staleness

PR #1024, opened 2024-06-17
feat: customizable `environment` in `GitHubActionStep`

a-bigelow:
`GitHubActionStep` may require access to variables/secrets only available within a GitHub [environment](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment). 
This PR adds `environment` as an optional input prop for `GitHubActionStep`, allowing users to select an environment per `GitHubActionStep` as they see fit.


kaizencc (changes requested):
I hate (and refuse to accept :)) an API that has both `env` and `environment` params. I know you are trying to differentiate in the documentation but thats not enough. 

Do you like `github_environment`? or `github_env_name`.

a-bigelow:
> I hate (and refuse to accept :)) an API that has both `env` and `environment` params. I know you are trying to differentiate in the documentation but thats not enough.
> 
> Do you like `github_environment`? or `github_env_name`.

`githubEnvironment` it is.

src/steps/github-action-step.ts
```diff
@@ -25,12 +33,14 @@ export interface GitHubActionStepProps {
 export class GitHubActionStep extends Step {
   public readonly env: Record<string, string>;
   public readonly jobSteps: JobStep[];
+  public readonly environment?: string;
```
kaizencc, line 36: lets make this githubEnvironment too. essentially i want _us_ to call this thing githubEnvironment everywhere internally

kaizencc (approved):

PR #1105, opened 2024-09-05
fix: upload-artifact v4 breaking change 

### LINKED ISSUE #1103: Breaking Change in GitHub actions/upload-artifact@v4
In GitHub's https://github.com/actions/upload-artifact/pull/598, they disabled the inclusion of dot files/hidden files by default without releasing a major version change.

As a result, the `cdk.out` uploaded by CDK Pipelines for GitHub Workflows now omit dot files. The missing dot files have begun breaking several of my pipelines that rely on `DockerImageFunction`s that reference dotfiles like `.yarnrc.yml` while building a container image asset because required files are missing from the staged assets.

We need the upload-artifact action input to include `include-hidden-files: true` so the hidden files we require are included. https://github.com/cdklabs/cdk-pipelines-github/blob/fa8d504166e3cf9d792b0a69708b395879467c54/src/pipeline.ts#L792-L796 

References:
- Issue: https://github.com/actions/upload-artifact/issues/602
- Their PR: https://github.com/actions/upload-artifact/pull/598

kaizencc:
fixes #1103


misterjoshua:
From the build error, it looks like the build artifact's uploaded with `include-hidden-files: false` and when packaging for each language, the `.projen` dir is missing, so there's no `.projen/tasks.json` and Projen can't find the language packaging tasks.

Projen appears to be tracking this issue as it's directly affected by GitHub's breaking change as well: https://github.com/projen/projen/issues/3820

src/pipeline.ts
```diff
@@ -920,6 +921,7 @@ export class GitHubWorkflow extends PipelineBase {
       with: {
         name: CDKOUT_ARTIFACT,
         path: dir,
+        ['include-hidden-files']: true,
```
mrgrain, line 924: This seems fine.

src/pipeline.ts
```diff
@@ -793,6 +793,7 @@ export class GitHubWorkflow extends PipelineBase {
         with: {
           name: output.fileSet.id,
           path: output.directory,
+          ['include-hidden-files']: true,
```
mrgrain, line 796: This I would check to see what we are actually including in the artifact and why.

mrgrain (approved):

PR #1451, opened 2026-03-30
fix: bump default GitHub Actions to latest major versions

zezke:
Bumps the hardcoded default action versions used in generated workflows:

  - `actions/checkout` v4 → v6
  - `actions/download-artifact` v4 → v8
  - `actions/upload-artifact` v4 → v7
  - `aws-actions/aws-cloudformation-github-deploy` v1 -> v2
  - `docker/login-action` v3 -> v4

GitHub is deprecating Node.js 16 and 20-based actions.

This fixes #1452

PR #1458, opened 2026-04-21
fix: bump configure-aws-credentials action to v6

zezke:
I actually forgot to include the `aws-actions/configure-aws-credentials` action in PR #1451.

This PR addresses that and bumps `aws-actions/configure-aws-credentials` to v6.


mrgrain (changes requested):
I think we should just upgrade it to v6. I don't see the need for this to be configurable.

zezke:
Fair point, it is a bit much to make this configurable. PR updated.

mrgrain (approved):

